PR #4, opened 2020-02-20
Register callbacks for return annotations

### LINKED ISSUE #3: Register callbacks on return types
`register_type` (or a new function) should also be allowed to register callbacks when that type is defined as a return annotation.  See https://github.com/napari/napari/pull/981#issuecomment-587988018

tlambert03:
closes #3  by adding a `return_callback` argument to `register_type`.  Multiple return callbacks can be registered for any given type. 

A side observation here:
the most natural way for an image arithmetic function to work would be for the function to return a numpy array, and _napari_ calls `add_image(array)`

```python
def image_arithmetic(layerA: Image, operation: Operation, layerB: Image) -> np.ndarray:
    return operation.value(layerA.data, layerB.data)
```

but that requires napari to register the type `np.ndarray` and assume they should _all_ be **Image** layers... which is a big assumption.  alternatively, we leave it to the user to return a napari layers, they have build it themselves and return an Image type... then it's easy for napari to know what to do... but the function becomes more napari specific though, I suppose, it already was by nature of accepting napari layers.

```python
def image_arithmetic(layerA: Image, operation: Operation, layerB: Image) -> Image:
    return Image(operation.value(layerA.data, layerB.data))
```

another possible solution is to define a new `LayerData` type which would essentially hand the return value to `_add_layer_from_data`.  But that's all a conversation for another day and doesn't really affect this PR.

### DIFF
--- a/examples/napari_image_arithmetic.py
+++ b/examples/napari_image_arithmetic.py
@@ -3,7 +3,7 @@ from enum import Enum
 
 import numpy
 
-from magicgui import magicgui
+from magicgui import magicgui, register_type
 from napari import Viewer, gui_qt, layers
 
 
@@ -33,6 +33,32 @@ class Operation(Enum):
 # If it is a callable, it will be called with the type of the function parameter being
 # rendered. In this case, we want any subclass of layers.Layer to be rendered as a
 # dropdown, but we only want layers of the type specified in the type hint to be shown.
+def get_layer_choices(gui, layer_type):
+    try:
+        # look for the parent Viewer based on where the magicgui is docked.
+        # if the magicgui widget does not have a parent, it is unattached
+        # to any viewers, and therefore we cannot return a list of layers
+        viewer = gui.parent().qt_viewer.viewer
+        return tuple(l for l in viewer.layers if isinstance(l, layer_type))
+    except AttributeError:
+        return ()
+
+
+def show_result(gui, result):
+    """Show result of image_arithmetic in viewer."""
+    try:
+        viewer = gui.parent().qt_viewer.viewer
+    except AttributeError:
+        viewer = None
+
+    if viewer and (result is not None):
+        try:
+            viewer.layers["result"].data = result
+        except KeyError:
+            viewer.add_image(data=result, name="result")
+
+
+register_type(layers.Layer, choices=get_layer_choices, return_callback=show_result)
 
 
 with gui_qt():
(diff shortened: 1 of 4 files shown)


jni:
@tlambert03 yeah, that's the basic idea behind the `image-types` repo: we need *separate* hierarchies of array-protocol types, one for images, another for points, another for vectors, etc. All of these are numpy arrays or duck arrays, but the type contains information about what kind of data the array holds. I'm still unclear about whether the Python type system can handle this, but I think it can via NewType, *assuming* NewType can be combined with protocols.

tlambert03:
yep, agreed re: `image-types`.
I'm not sure we need to combine NewType with protocols though.  I'm assuming you'd be starting from `AnyImage = NewType('AnyImage', np.ndarray)` right?  I think that would be useful if we wanted everything to be a subtype of np.ndarray... whereas protocols need only check various methods and attributes.  
I had some success just using Protocols recently, but it required me to first install numpy-stubs... and I'm not sure there's going to be any way around that: due to numpy's c libraries, I don't think Protocols needs some sort of stub even to check that the methods exist.

jni:
What I meant by combine is `AnyImage = NewType('AnyImage', ArrayProtocol)`. ...? 😬 

I don't have a huge problem with depending on numpy-stubs *assuming it fits all our needs*. Otherwise we can just rip out the bits we want...

tlambert03:
`mypy` doesn't like it:
```python
from typing import NewType, Protocol, Tuple

class HasShape(Protocol):
    shape: Tuple[int, ...]

Shapey = NewType("Shapey", HasShape)
# Error: "NewType cannot be used with protocol classes"
```
i'm still picking all this up... but what is the need for the `NewType` there at all?  I imaged doing something like

class HasShape(Protocol):
    shape: Tuple[int, ...]

class HasShape(Protocol):
    shape: Tuple[int, ...]

jni:
Because how can we express the difference between a function that returns an array of coordinates vs one that returns an image?

tlambert03:
I _think_ we can define an intersection of protocols with multiple inheritance, and use [typing.Literal](https://docs.python.org/3/library/typing.html#typing.Literal) to restrict to certain explicit shapes.

```python
class Is2D(Protocol):
    shape: Tuple[int, int]
    ndim: Literal[2]

class Array2D(ArrayLike, Is2D, Protocol):  ...
```

jni:
But even a 2D array can equally be an image or a set of coordinates!

tlambert03:
well then we're forked. ;)
ok good point... gotta keep thinking!

jni:
We need to write a comprehensive writeup to put it to the mypy site. It's unclear that this *shouldn't* work. Otherwise we might end up writing a PEP... =D

tlambert03:
FWIW, this is mentioned in pep 544 https://www.python.org/dev/peps/pep-0544/#newtype-and-type-aliases

jni:
Hmmm, that is super vague LOL

PR #14, opened 2020-05-18
Initial support for labels

tlambert03:
This PR adds labels (using the argument name) by default.  Can be turned off with `@magicgui(labels=False)`... works with both `@magicgui(layout='horizontal')` and `'vertical'`
Screen Shot 2020-05-18 at 2 13 40 PM
Screen Shot 2020-05-18 at 2 13 58 PM


jni:
:heart_eyes:

PR #22, opened 2020-06-22
Add datetime to type2widget function

GenevieveBuckley:
magicgui has support for date time choosers (with PyQt QDateTimeEdit), but there is no automatic inference if you pass in a datetime object. Since `datetime` is part of the standard library, I think it would be nice to add this to the `type2widget` function.

### DIFF
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -3,6 +3,7 @@
 
 import sys
 from contextlib import contextmanager
+import datetime
 from enum import Enum, EnumMeta
 from typing import Any, Callable, Dict, Iterable, NamedTuple, Optional, Tuple, Type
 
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
I'm not sure why only the linux tests are failing - I don't think that has anything to do with what I've done in this PR

tlambert03:
linux tests are fixed now (was a qt 5.15 thing).  This looks great thanks!
Can we add one more thing?  for the getter, let's convert back to the native python type (instead of a `QDateTime`).  Looks like qtpy isn't handling this perfectly, so use this as the getter in the `getter_setter_onchange` for `QDateTimeEdit`:
```python
        def getter():
            try:
                return widget.dateTime().toPython()
            except TypeError:
                return widget.dateTime().toPyDateTime()

```

GenevieveBuckley:
Are you willing to overlook the two lines of uncovered code here? I think that's the only thing stopping this PR going in.

The two uncovered lines reported by codecov are the except block here:
```
        def getter():
            try:
                return widget.dateTime().toPython()
            except TypeError:
                return widget.dateTime().toPyDateTime()
```

I'm not sure how to create a type error in `widget.dateTime().toPython()` and force the except clause to run. I think whether you need `toPython()` or `toPyDateTime()` depends on your system (my local dev environment uses the except handling block and works great).

Issue #20, opened 2020-06-04
Feature request: support for QFileDialog (file and directory choosers)


GenevieveBuckley:
One very common component of GUIs I build for users is a file or directory chooser. The use case is usually:
* select the input data (might be a folder full or images, or sometimes a single input file)
* select a folder where you want to save the output results
* tell me what other parameters you want to use for the analysis

There is [QFileDialog](https://doc.qt.io/qt-5/qfiledialog.html), which I think would fulfill this use case.

tlambert03:
That’s sounds good.  What should the argument type be for magicgui to render a dialog like this?  Maybe pathlib.Path?

```python
@magicgui
def func(path: pathlib.Path):
    ...
```

tlambert03:
@GenevieveBuckley ... any chance you want to get your hands dirty and submit a PR for this?  I can definitely direct you to the appropriate places

tlambert03:
here's roughly what needs to be done (all in `magicgui/_qt.py`):
- add a mapping from `pathlib.Path` to `QFileDialog` in the `type2widget` function
- add an `elif isinstance(widget, QFileDialog):` section to `getter_setter_onchange`, and then choose which methods from QFiledialog to use as the "getter", "setter", and "on_change" signal. (follow the pattern for the others)
- lastly, we'll probably need to make our own QFileDialog subclass that has some things set to make it work inside other widgets... minimally something like:

```python
from qtpy.QtCore import Qt, QFileDialog

class MagicFileDialog(QFileDialog):
    def __init__(self, parent=None):
        super().__init__(parent)
        self.setOption(QFileDialog.DontUseNativeDialog)
        self.setWindowFlags(self.windowFlags() & ~Qt.Dialog)
```
(and then, actually, use `MagicFileDialog`  in the `type2widget` map.

we may want to further customize that FileDialog once we play with it a bit (for instance... but removing the open button?)

GenevieveBuckley:
Thanks Talley! I had a small play around with this last night, I think I need to have more of a look at the getter/setter pattern you have for the other widgets. I'm going to be busy for the next few days but I'll let you know how I go

VolkerH:
Just wondered whether magicgui supports `pathlib.Path` types in combination with file dialogs and the search turned up this issue. 
Just curios how one could enforce:

- selecting a single file only
- selecting multiple files
- selecting a single folder only
- selecting either folder or file

Does it make sense to declare subtypes for this or do any of you have a better idea for this?

[...]

PR #23, opened 2020-06-23
Filedialog widget for magicgui

GenevieveBuckley:
This is for discussion of https://github.com/napari/magicgui/issues/20 QFileDialog picker.

Useful to know:
* It's not enough to simply map from a Path type to a QFileDialog/MagicFileDialog widget in `type2widget`, we need to check `isinstance` instead. This takes care of platform specific things like WindowsPath, etc.

Still up for discussion:
* I think it'd be nicer to have a button & text label that you can push to launch the file dialog, rather than keeping the file dialog expanded in magicgui.
* The getter/setter for the file dialog here is only for files, we still need to work out what to do about files vs directories vs multiple selection, etc.

Here's a
```python
"""FileDialog with magicgui."""
from pathlib import Path
from magicgui import event_loop, magicgui

@magicgui(call_button='Call Button')
def filepicker(filename=Path()):
    print("The filename is:", filename)
    return filename

with event_loop():
    gui = filepicker.Gui(show=True)
    gui.filename_changed.connect(print)
    # we can connect a callback function to the __call__ event on the function
    gui.called.connect(lambda x: gui.set_widget("result", str(x), position=-1))
```

File dialog GUI is embedded
image

Embedded GUI disappears completely once you select a file
image

Pairing file dialog with a version of the call button might be better, this launches the file picker when you need it and doesn't disappear
image

### DIFF
--- a/examples/file_dialog.py
+++ b/examples/file_dialog.py
@@ -1,0 +1,17 @@
+"""FileDialog with magicgui."""
+from pathlib import Path
+from magicgui import event_loop, magicgui
+
+
+# may also add Qt-style filter to filename options:
+# e.g. {"filter": "Images (*.tif *.tiff)"}
+@magicgui(filename={"mode": "existing_file"})
+def filepicker(filename=Path("~")):
+    """Take a filename and do something with it."""
+    print("The filename is:", filename)
+    return filename
+
+
+with event_loop():
+    gui = filepicker.Gui(show=True)
+    gui.filename_changed.connect(print)
--- a/examples/file_dialog.py
+++ b/examples/file_dialog.py
@@ -5,7 +5,7 @@ from magicgui import event_loop, magicgui
 
 # may also add Qt-style filter to filename options:
 # e.g. {"filter": "Images (*.tif *.tiff)"}
-@magicgui(filename={"mode": "existing_file"})
+@magicgui(filename={"mode": "r"})
 def filepicker(filename=Path("~")):
     """Take a filename and do something with it."""
     print("The filename is:", filename)
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -215,6 +215,10 @@ def getter_setter_onchange(widget: WidgetType) -> GetSetOnChange:
         )
     elif isinstance(widget, QSplitter):
         return GetSetOnChange(widget.sizes, widget.setSizes, widget.splitterMoved)
+    elif isinstance(widget, MagicFileDialog):
+        return GetSetOnChange(
+
+        )
     raise ValueError(f"Unrecognized widget Type: {widget}")
 
 
@@ -308,3 +312,12 @@ class QDoubleSlider(QSlider):
     def setMaximum(self, value):
         """Set maximum position of slider in float units."""
         super().setMaximum(value * self.PRECISION)
+
+
+class MagicFileDialog(QFileDialog):
+    """A FileDialog widget that is launched by a button press."""
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+        self.setOption(QFileDialog.DontUseNativeDialog)
+        self.setWindowFlags(self.windowFlags() & ~Qt.Dialog)
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -431,7 +431,7 @@ class MagicFileDialog(QWidget):
         """Get current file path."""
         text = self.line_edit.text()
         if self.mode is FileDialogMode.EXISTING_FILES:
-            return tuple(Path(p) for p in text.split(","))
+            return tuple(Path(p) for p in text.split(", "))
         return Path(text)
 
     def set_path(self, value: Union[List[str], Tuple[str, ...], str, Path]):
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -349,12 +349,18 @@ class FileDialogMode(Enum):
     EXISTING_FILES - return one or more existing files.
     OPTIONAL_FILE - return one file name that does not have to exist.
     EXISTING_DIRECTORY - returns one existing directory.
+    R - read mode, returns one existing file. Alias of EXISTING_FILE.
+    W - write mode, returns one file name that does not have to exist.
+        Alias of OPTIONAL_FILE.
     """
 
     EXISTING_FILE = "getOpenFileName"
     EXISTING_FILES = "getOpenFileNames"
     OPTIONAL_FILE = "getSaveFileName"
     EXISTING_DIRECTORY = "getExistingDirectory"
+    # Aliases
+    R = "getOpenFileName"
+    W = "getSaveFileName"
 
 
 class MagicFileDialog(QWidget):
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -404,7 +404,10 @@ class MagicFileDialog(QWidget):
                 try:
                     mode = FileDialogMode[value.upper()]
                 except KeyError:
-                    pass
+                    pass  # leave mode as string type, raises ValueError later
+        # If mode is not a valid FileDialogMode enum type (eg: input string
+        # could not be recognised and converted properly in the if block above)
+        # then we raise a ValueError to alert the user.
         if not isinstance(mode, FileDialogMode):
             raise ValueError(
                 f"{mode!r} is not a valid FileDialogMode. "
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -349,7 +349,8 @@ class FileDialogMode(Enum):
     EXISTING_FILES - return one or more existing files.
     OPTIONAL_FILE - return one file name that does not have to exist.
     EXISTING_DIRECTORY - returns one existing directory.
-    R - read mode, returns one existing file. Alias of EXISTING_FILE.
+    R - read mode, returns one or more existing files.
+        Alias of EXISTING_FILES.
     W - write mode, returns one file name that does not have to exist.
         Alias of OPTIONAL_FILE.
     """
@@ -359,7 +360,7 @@ class FileDialogMode(Enum):
     OPTIONAL_FILE = "getSaveFileName"
     EXISTING_DIRECTORY = "getExistingDirectory"
     # Aliases
-    R = "getOpenFileName"
+    R = "getOpenFileNames"
     W = "getSaveFileName"
 
 
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -360,7 +360,7 @@ class FileDialogMode(Enum):
     OPTIONAL_FILE = "getSaveFileName"
     EXISTING_DIRECTORY = "getExistingDirectory"
     # Aliases
-    R = "getOpenFileNames"
+    R = "getOpenFileName"
     W = "getSaveFileName"
 
 
--- a/magicgui/_qt.py
+++ b/magicgui/_qt.py
@@ -4,6 +4,7 @@
 import sys
 from contextlib import contextmanager
 from enum import Enum, EnumMeta
+import pathlib
 from typing import Any, Callable, Dict, Iterable, NamedTuple, Optional, Tuple, Type
 
 from qtpy.QtCore import Qt, Signal
@@ -16,6 +17,7 @@ from qtpy.QtWidgets import (
     QComboBox,
     QDateTimeEdit,
     QDoubleSpinBox,
+    QFileDialog,
     QFormLayout,
     QGridLayout,
     QGroupBox,
@@ -152,6 +154,7 @@ def type2widget(type_: type) -> Optional[Type[WidgetType]]:
         int: QSpinBox,
         float: QDoubleSpinBox,
         str: QLineEdit,
+        pathlib.Path: QFileDialog,
         type(None): QLineEdit,
     }
     if type_ in simple:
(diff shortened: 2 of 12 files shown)


tlambert03:
> I think it'd be nicer to have a button & text label that you can push to launch the file dialog, rather than keeping the file dialog expanded in magicgui.

this is totally how this should be,  great point!  makes everything easier (and let's us use the native file dialogs).  So we basically just need to make our own new widget that has a LineEdit and a button.  I'll propose some changes that you can review.

GenevieveBuckley:
I've merged your suggestion, so this is ready for review now. It still needs a couple of tests added.

I think we should stick with the class name `MagicFileDialog` instead of the less clear `MagicPathLineEdit`, but other than that I love it

GenevieveBuckley:
I'm not sure how to properly test the `_on_choose_clicked()` method without the newly opened file dialog box blocking until it is manually closed. I did see [this suggestion](https://github.com/pytest-dev/pytest-qt/issues/256#issuecomment-477816887) to use QTimer, but I don't know how to get the handle/reference of the spawned file dialog window we want to close.

examples/file_dialog.py
```diff
@@ -0,0 +1,17 @@
+"""FileDialog with magicgui."""
+from pathlib import Path
+from magicgui import event_loop, magicgui
+
+
+# may also add Qt-style filter to filename options:
+# e.g. {"filter": "Images (*.tif *.tiff)"}
+@magicgui(filename={"mode": "existing_file"})
```
jni, line 8: Didn't we agree to use 'mode': 'r' for this?

jni: Other than that the code looks really good! Like that's it, that's my full review! :joy:

tlambert03: yeah I liked that, @GenevieveBuckley, can you make that change?  and instead of "files", make it a `List[Path]` type hint?  might run into some issues there, but let me know

tlambert03:
> I'm not sure how to properly test the `_on_choose_clicked()` method without the newly opened file dialog box blocking until it is manually closed. I did see [this suggestion](https://github.com/pytest-dev/pytest-qt/issues/256#issuecomment-477816887) to use QTimer, but I don't know how to get the handle/reference of the spawned file dialog window we want to close.

yeah, that _is_ a tricky one!  i'm not immediately sure... might need to play around with that.  but one immediate thought is that you might be able to use the QTimer pattern here, and then search `QApplication.topLevelWidgets()` for an instance of `QFileDialog`?  alternatively... we can just instantiate (and save) the QFileDialog in our `_on_choose_clicked` method, and then call exec_ manually.  but that seems unfortunate just for testing.

tlambert03:
this sort of approach will get you a handle on the dialog during a test:
```python
    def handle_dialog():
        dialog = next(
            child
            for child in filewidget.children()
            if isinstance(child, QFileDialog)
        )
        print(dialog)
        dialog.reject()  #still hangs

    QtCore.QTimer().singleShot(400, handle_dialog)
    filewidget._on_choose_clicked()
```
but I still can't _close_ it.  It might be necessary to move away from the pattern of using [static methods](https://doc.qt.io/qt-5/qfiledialog.html#static-public-members) as I did, and instead make a filedialog, use [setFileMode](https://doc.qt.io/qt-5/qfiledialog.html#fileMode-prop), then call `.exec_()`

GenevieveBuckley:
Oh that's really weird @tlambert03 - your suggestion ONLY hangs on Mac.

examples/file_dialog.py
GenevieveBuckley: I'm sorry @tlambert03, what do you mean by "instead of "files", make it a List[Path] type hint?". Where in the code should I be looking?

GenevieveBuckley: @jni I'd prefer not to replace Talley's names, but instead add `'r'` and '`w'` as aliases (because my brain finds it confusing to remember one mixed set of naming conventions but is a-ok with having an optional alternative way to specify the same thing)

```python
class FileDialogMode(Enum):
    """FileDialog mode options.

    EXISTING_FILE - returns one existing file.
    EXISTING_FILES - return one or more existing files.
    OPTIONAL_FILE - return one file name that does not have to exist.
    EXISTING_DIRECTORY - returns one existing directory.

    r - read mode, returns one existing file. Alias of EXISTING_FILE.
    w - write mode, returns one file name that does not have to exist.
        Alias of OPTIONAL_FILE.
    """

    EXISTING_FILE = "getOpenFileName"
    EXISTING_FILES = "getOpenFileNames"
    OPTIONAL_FILE = "getSaveFileName"
    EXISTING_DIRECTORY = "getExistingDirectory"
    # Aliases
    r = "getOpenFileName"
    w = "getSaveFileName"
```

jni: I'm ok with aliases, but (a) I would definitely prefer for 'r' to be the "promoted" use, and (b) I'd go for `r='getOpenFileNames'`. btw, does that allow selecting of directories? I think it does iirc.

GenevieveBuckley: I'm happy for 'r'/'w' to be the promoted use too @jni

Sadly no, `'getOpenFileNames'` does not allow selecting directories. There isn't an option available that will do that.

magicgui/_qt.py
```diff
@@ -312,3 +336,106 @@ def setValue(self, value):
...
+    def mode(self, value: Union[FileDialogMode, str]):
+        mode: Union[FileDialogMode, str] = value
+        if isinstance(value, str):
+            try:
+                mode = FileDialogMode(value)
+            except ValueError:
+                try:
+                    mode = FileDialogMode[value.upper()]
+                except KeyError:
+                    pass
```
GenevieveBuckley, line 403: Shouldn't this raise a helpful error message rather than passing silently?

GenevieveBuckley: I also can't figure out how to trigger this logic branch, in what circumstance does it happen?

GenevieveBuckley:
I'm very happy with this overall! I'd like to get it in soon.

To summarize where we're at:

1. I didn't understand [this comment](https://github.com/napari/magicgui/pull/23/#discussion_r448526189), so I'm not sure what to do here:
> instead of "files", make it a List[Path] type hint? might run into some issues there, but let me know

2. I'm a hair under the goal test coverage, but I'm not sure how to improve that at this point.

3. I'm not sure how to trigger the mode setter `except KeyError` clause for MagicFileDialog. Maybe this doesn't really matter, I noticed because I was looking for ways to bump the test coverage.

magicgui/_tests/test_ qt.py
```diff
@@ -90,3 +101,75 @@ def test_set_categorical(qtbot):
...
+
+    # set the mode
+    filewidget.mode = _qt.FileDialogMode.EXISTING_FILES  # Enum input
+    assert filewidget.mode == _qt.FileDialogMode.EXISTING_FILES
+    filewidget.mode = "oPtioNal_FiLe"  # improper capitalization
+    assert filewidget.mode == _qt.FileDialogMode.OPTIONAL_FILE
+    filewidget.mode = "EXISTING_DIRECTORY"  # string input
+    assert filewidget.mode == _qt.FileDialogMode.EXISTING_DIRECTORY
+    with pytest.raises(ValueError):
+        filewidget.mode = 123  # invalid mode
```
tlambert03, line 122:
```suggestion
    with pytest.raises(ValueError):
        filewidget.mode = 123  # invalid mode
    with pytest.raises(ValueError):
        filewidget.mode = 'invalid_string'
```
(need to check for an invalid string... not just an invalid type)

magicgui/_qt.py
tlambert03: It will happen if a string is passed that is not a valid choice.  First it checks the enum values, then it checks the enum names.  If neither of those pass, then `mode` will _still_ be a string and then next `ValueError` will be raised (the `pass` let's us write a single `ValueError`).  I added a suggestion for how to test it, by testing an invalid string, not just an invalid type `123`)

tlambert03:
> 1. I didn't understand [this comment](https://github.com/napari/magicgui/pull/23/#discussion_r448526189), so I'm not sure what to do here:
> 
> > instead of "files", make it a List[Path] type hint? might run into some issues there, but let me know

That was in reference to @jni's https://github.com/napari/magicgui/issues/20#issuecomment-648504543 where he suggested that "it might be better to go with r/w for modes (existing file/new file), and Sequence[Path] for existing files."  In other words, if the user wants to allow for selection of multiple files, they should annotate the argument as a `Sequence` of `Path` objects (and then we will render the appropriate widget).  However, that would mean we can no longer _just_ look at the options dict... so it does get a bit trickier.  I can take a look later

magicgui/_tests/test_ qt.py
```diff
@@ -90,3 +101,75 @@ def test_set_categorical(qtbot):
...
+    assert w is not None
+    assert _qt.MagicFileDialog in w.mro()  # since type(w) = sip.wrappertype
+
+
+@pytest.mark.skipif("sys.platform != 'windows'")  # Linux and Mac test
+def test_linux_mac_magifiledialog(qtbot):
+    """Test we get a MagicFileDialog from a PosixPath type."""
+    w = _qt.type2widget(PosixPath)
+    assert w is not None
+    assert _qt.MagicFileDialog in w.mro()  # since type(w) = sip.wrappertype
```
tlambert03, line 175: since we're not actually instantiating any of these classes, we can run this test on all pathtypes on all platforms.  also, `type2widget` is actually kinda poorly named :) it should really be `type2widgetclass` or something.  So we can actually just test for equality here.

```suggestion
@pytest.mark.parametrize("pathtype", [PosixPath, WindowsPath, Path])
def test_linux_mac_magifiledialog(pathtype):
    """Test we get a MagicFileDialog from a various Path types."""
    assert _qt.type2widget(pathtype) == _qt.MagicFileDialog
```

GenevieveBuckley: This is beautiful and concise, I like your change a lot!

magicgui/_tests/test_ qt.py
```diff
@@ -90,3 +101,66 @@ def test_set_categorical(qtbot):
...
+    filewidget.set_path(['path/3.txt, path/4.txt'])
+    assert filewidget.get_path() == (Path('path/3.txt'), Path('path/4.txt'))
+
+    with pytest.raises(TypeError):
+        filewidget.set_path(123)  # invalid type, only str/Path accepted
+
+
+@pytest.mark.skipif("sys.platform == 'darwin'")  # dialog box hangs on Mac
+@pytest.mark.parametrize("mode", [
+    ('r'),  # existing file
```
tlambert03, line 141: this actually isn't surviving black formatting.
https://github.com/napari/magicgui/pull/23/checks?check_run_id=829849809#step:5:15

but it did reveal an issue in my linting test! it's (uselessly) reformatting files without raising an error. fixing in #26, but maybe try to run black locally on this?

magicgui/_qt.py
GenevieveBuckley: Thanks!
I think the source of my confusion was assuming the second `if` block was equivalent to an `elif` when I first skimmed over the code. Perhaps if I'd revisited that assumption later I would have caught my own error in logic.

tlambert03: yeah, it's also a rather implicitly coded block.  might be nice to add a comment there

tlambert03:
regarding using `Sequence[Path]`... I'm fine leaving that for another PR. we might need to touch a some additional code to make that work.

magicgui/_tests/test_ qt.py
```diff
@@ -90,3 +101,63 @@ def test_set_categorical(qtbot):
...
+    assert filewidget.get_path() == (Path("path/one.txt"), Path("path/two.txt"))
+    filewidget.set_path(["path/3.txt, path/4.txt"])
+    assert filewidget.get_path() == (Path("path/3.txt"), Path("path/4.txt"))
+
+    with pytest.raises(TypeError):
+        filewidget.set_path(123)  # invalid type, only str/Path accepted
+
+
+@pytest.mark.skipif("sys.platform == 'darwin'")  # dialog box hangs on Mac
+@pytest.mark.parametrize("mode", [("r"), ("EXISTING_DIRECTORY")])
```
tlambert03, line 140:
```suggestion
@pytest.mark.parametrize("mode", ["r", "EXISTING_DIRECTORY"])
```
?

magicgui/_qt.py
```diff
@@ -312,3 +340,106 @@ def setValue(self, value):
...
+    W - write mode, returns one file name that does not have to exist.
+        Alias of OPTIONAL_FILE.
+    """
+
+    EXISTING_FILE = "getOpenFileName"
+    EXISTING_FILES = "getOpenFileNames"
+    OPTIONAL_FILE = "getSaveFileName"
+    EXISTING_DIRECTORY = "getExistingDirectory"
+    # Aliases
+    R = "getOpenFileNames"
```
tlambert03, line 363:
```suggestion
    R = "getOpenFileName"
```

let's make this an alias for a single file.  and in a followup (or here if you want to) we can do something like this:
```python
@magicgui
def filepicker(filename=Sequence[Path("~")]):
    ...
```
and then in `type2widget` return `functools.partial(MagicFileDialog, mode='EXISTING_FILES')`

GenevieveBuckley:
> 
> 
> regarding using `Sequence[Path]`... I'm fine leaving that for another PR. we might need to touch a some additional code to make that work.

That's my feeling too, glad it's not just me feeling lazy

magicgui/_qt.py
tlambert03: whoops, sorry. I committed this and it broke the test.  can you fix that?  Then I think this is good to go

GenevieveBuckley: Ok

tlambert03:
thanks!

PR #31, opened 2020-07-13
add literal eval fallback widget

### LINKED ISSUE #29: Use eval'd text box as fallback when type inference fails
Many magic typing failures happen for simple arguments (tuple, range, list, (small) dict...) that could actually be easy for users to type in. Rather than fail, magicgui could present a text box that gets `eval`'d.

tlambert03:
closes #29 by adding a `LiteralEvalEdit ` that is used when type inference fails.  As this uses `literal_eval` instead of `eval`, this is a "safe" fix, but will miss a lot of edge cases.


magicgui/core.py
```diff
@@ -326,14 +326,20 @@ class instance using setattr(self, self.WIDGET_ATTR.format(name), widget).
             try:
                 WidgetType = type2widget(arg_type)
             except TypeError:
+                funcname = getattr(self.func, "__name__", repr(self.func))
```
tlambert03, line 329: not directly related to this PR ... but I found that `partial` functions weren't working since they don't have a `__name__` or `__annotation__` attribute.

PR #35, opened 2020-07-20
Colorpicker widget

GenevieveBuckley:
This PR adds a [colorpicker widget](https://doc.qt.io/qt-5/qcolor.html) to magicgui, mapped to this [Color type](https://github.com/vaab/colour) (third party library [available on PyPI](https://pypi.org/project/colour/)). This PR doesn't actually add a dependency requirement to magicgui, but you don't have it installed then you don't get access to the colorpicker widget.

I started this branch a while ago when I was getting familiar with the magicgui concepts I needed for the filepicker dialog, and I think it'd be a fun addition.

This still needs tests, but I wanted to get your overall thoughts before pushing ahead.


tlambert03:
Great idea!  With the color picker being a dialog though, do you think we should do something like the file dialog (where it's a LineEdit and you have to click a button to get the dialog)

Maybe add a simple example so we can play with it?

tlambert03:
Another thought: that line edit could have an autocomplete when typing color names, and a little patch that can be clicked that shows the actual QColor dialog. See the napari color widget for examples (though here i think we should stick with the native color picker if possible)

GenevieveBuckley:
Closing this. If we pick this back up in the future, there's two things to pay attention to:
1. Use a line edit + color patch + button to launch colorpicker
2. Use whatever the resolution to https://github.com/napari/magicgui/issues/40 is (about a nice way to handle potentially supported third party libraries that aren't required dependencies of magicgui)

PR #36, opened 2020-07-21
Widget demo example script

GenevieveBuckley:
I thought it would be good to have an easy to run example script showcasing all the widget types supported by `magicgui`. 

Right now there's nowhere in the docs or examples that explains all the different types of widgets you can have. I personally learnt most of what I know about the [Gooey](https://github.com/chriskiehl/Gooey) from [their widget demo script](https://github.com/chriskiehl/GooeyExamples/blob/master/examples/widget_demo.py), so that's a big motivation to include something similar here.

### DIFF
--- a/examples/widget_demo.py
+++ b/examples/widget_demo.py
@@ -3,8 +3,10 @@
 from enum import Enum
 from datetime import datetime
 import math
+from pathlib import Path
 
 from magicgui import event_loop, magicgui
+
 
 class Medium(Enum):
     """Enum for various media and their refractive indices."""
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
> love it! I was gonna say add a slider... but I think we need to have our own API for that (rather than using "widget_type: QSlider"). So lets' add that once we have it.

I forgot about sliders! I'm not sure what you mean by "we need to have our own API... let's add that once we have it", can you elaborate? Ideally I'd like to include as many things as possible into the examples.

tlambert03:
I mean that every feature should have some “magicgui” representation/API that we wouldn’t mind reusing even if we switched to another backend.  So we should be able to do `widget_type=slider` where “slider” means something special in magicgui.  Not for this PR of course...  just that I’ve never liked how you’re supposed to ask for a slider.  (I imagine there are other cases too, where there are multiple suitable widgets for a given type)

tlambert03:
I'm going to merge this, we can continue to embellish this example as the "showroom" script as we add more widgets and/or simplify the API.  thanks @GenevieveBuckley

PR #38, opened 2020-07-21
Add a logarithmic scale slider class

GenevieveBuckley:
Juan originally suggested a log scale slider for napari (https://github.com/napari/napari/issues/938) but this strikes me as something that would also be very useful to have available in `magicgui`.

Adds numpy as a base requirement.

The main limitation of the log slider is that the available range is restricted from 0 -> 14.5. If you go any higher than 14.5 you get an overflow error from Qt. Qt expects internal values in the range of -2147483647 to +2147483647, so the exponential can cause problems quickly. Still, I think there is some value in having a log slider available, even if there are some restrictions here.


magicgui/_qt/widgets/log_slider.py
```diff
@@ -0,0 +1,31 @@
...
+        """Set integer slier position from float ``value``."""
+        super().setValue(np.exp(value) * self.PRECISION)
+
+    def setMinimum(self, value):
+        """Set minimum position of slider in float units."""
+        super().setMinimum(np.exp(value) * self.PRECISION)
+
+    def setMaximum(self, value):
+        """Set maximum position of slider in float units."""
+        super().setMaximum(np.exp(value) * self.PRECISION)
```
jni, line 31: You could use the built-in `math` module for this instead of NumPy, so I don't think it's worth picking up the dependency just for these. (The test is also simple enough that you could implement it in pure Python)

GenevieveBuckley: Oh good point, will do

GenevieveBuckley:
Ok, something weird is going on with the tests. I can't see how the changes here could possibly have caused this, but here's what I've found.

* Tests pass on the current master branch (I can reproduce this locally).
* On this `log-slider` branch, all the CI test runs report a test failure for `magicgui/_tests/test_qt.py::test_event` (I can reproduce this locally too).
* If you run that specific test by itself, it passes fine: 'pytest magicgui/_tests/test_qt.py::test_event`
* But if you run any other test using a `qtbot` fixture (or the `magic_widget` fixture, since that fixture also uses `qtbot`) before it, `test_event` will fail because of the line: `assert not QtW.QApplication.instance()`. I've checked this theory by running an example test before `test_event` that does nothing but `assert True` but *does* take in the `qtbot` fixture:
```
def test_example(qtbot):
    assert True
```

So, what to do? 

If we can remove the line `assert not QtW.QApplication.instance()` from `test_event` then we'd have no more test failures - but I assume the reason it's there in the first place is because it's checking something important? I don't know why it's important that we see/expect a null pointer here (which is what the [Qt docs say is supposed to happen if no instance has been allocated](https://doc.qt.io/qt-5/qcoreapplication.html#instance)), so I'd really like to hear suggestions.

tlambert03:
that was a pretty fragile test to begin with and was very dependent on the order of operations.  Once a qtbot makes an app, it sticks around for the rest of the tests.  I was asserting that there was no instance so that we could make sure that the `event_loop` context manager (same as `napari.gui_qt` basically) was actually creating the app.

I just pushed a fix here, which deletes any pre-existing application instance using `sip.delete` or `shiboken.2` delete... a clever little trick that Phillip recently found on stack overflow.

GenevieveBuckley:
Thank you for the fix! Very helpful :)

tlambert03:
welp... looks like pyside2 was ok with that, but not pyqt5.  I just skipped the test for pyqt5.

magicgui/_qt/widgets/log_slider.py
```diff
@@ -0,0 +1,32 @@
...
+        self.setMinimum(0)
+        self.setMaximum(10)
+
+    def value(self):
+        """Get the natural log slider value as a float."""
+        return math.log(super().value() / self.PRECISION)
+
+    def setValue(self, value):
+        """Set integer slier position from float ``value``."""
+        super().setValue(math.exp(value) * self.PRECISION)
```
tlambert03, line 24:
```suggestion
        super().setValue(int(math.exp(value) * self.PRECISION))
```
getting new warnings in the tests: `Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.`

... and actually can you do the same to  `QDoubleSlider.setValue` and `setMaximum` while we're at it?

magicgui/_qt/widgets/log_slider.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def setMinimum(self, value):
+        """Set minimum position of slider in float units."""
+        super().setMinimum(math.exp(value) * self.PRECISION)
```
tlambert03, line 28:
```suggestion
        super().setMinimum(int(math.exp(value) * self.PRECISION))
```

magicgui/_qt/widgets/log_slider.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def setMaximum(self, value):
+        """Set maximum position of slider in float units."""
+        super().setMaximum(math.exp(value) * self.PRECISION)
```
tlambert03, line 32:
```suggestion
        super().setMaximum(int(math.exp(value) * self.PRECISION))
```

magicgui/_qt/widgets/log_slider.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, parent=None):
+        super().__init__(Qt.Horizontal, parent=parent)
+        self.setMinimum(0)
+        self.setMaximum(10)
+
+    def value(self):
+        """Get the natural log slider value as a float."""
+        return math.log(super().value() / self.PRECISION)
+
+    def setValue(self, value):
```
tlambert03, line 22: I'm about to enable stricter mypy errors, can you type hint these methods?

magicgui/_qt/widgets/log_slider.py
```diff
@@ -0,0 +1,32 @@
...
+class QLogSlider(QSlider):
+    """A Slider Widget with a natural logarithmic scale."""
+
+    PRECISION = 1000
+
+    def __init__(self, parent=None):
+        super().__init__(Qt.Horizontal, parent=parent)
+        self.setMinimum(0)
+        self.setMaximum(10)
+
```
tlambert03, line 17: to make this slightly more flexible, perhaps we should add an attribute `self.base = math.e` ... then convert `math.exp(value)` to `math.pow(self.base, value)` and `math.log(value)` to `math.log(value, base=self.base)`

GenevieveBuckley: Nice touch, I like this suggestion a lot. Done!

GenevieveBuckley:
Ok I think I've addressed all the reviewer suggestions, and we have a passing test suite now (thanks to Talley). I think things are looking good here.

PR #41, opened 2020-07-23
Add (slightly) strict mypy checking

tlambert03:
This PR would enable some of the stricter mypy options.  (most notably, `disallow_untyped_defs` which [disallows defining functions without type annotations or with incomplete type annotations.](https://mypy.readthedocs.io/en/stable/config_file.html#untyped-definitions-and-calls).  I'm not married to this, so if @jni, @GenevieveBuckley, or anyone else hates it, we can back off on it.  Probably the most notable consequence of this is that all function defs must have types, including return annotations (including `-> None:`).  If people hate that, I can turn off `disallow_untyped_defs`.

Partially... it's an experiment to add strict typing to a young project :)


GenevieveBuckley:
I don't know if I'd hate it or not, personally I haven't worked on any projects with strict mypy checking before so I don't know if I'd find it irritating or not. So I'm happy to try it and find out, if other people want to. Worst case scenario, we can always return to being less strict with mypy later on.

> Partially... it's an experiment to add strict typing to a young project :)

While a project is young is probably the best time, it's always a bit more of an effort later on.

>  Probably the most notable consequence of this is that all function defs must have types, including return annotations (including `-> None:`)

Does this mean someone needs to go through the code and add those? I saw class methods in QDoubleSlider/QLogSlider that don't return anything (so implicitly return None) don't have annotated return types. (That was just the most recent thing I looked at, I don't know what level of work needs to go into making all of magicgui compliant with your proposed change).

tlambert03:
> Worst case scenario, we can always return to being less strict with mypy later on.

that's my thought, we can always turn it off

> Does this mean someone needs to go through the code and add those

all the fixes are included in the current PR, so it passes now.  it's also in the pre-commit hook and I have "pretty" output turned on, so you'd see something like this if you had an incomplete def:

```pytb
magicgui/core.py:728:1: error: Function is missing a return type annotation  [no-untyped-def]
    def reset_type(type_: Type):
    ^
magicgui/core.py:735:1: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
    def _type2choices(type_) -> ChoicesType:
    ^

```

jni:
>  (including -> None:)

I do hate this...

Imma let you drive on this Talley but I'm blocking my eyes. 🙈 😝

tlambert03:
That's why I asked! I don't care too much.  Mostly I just wanted a way to make sure that the function arguments were annotated, but it looks like there's no config option to just allow the return type to be missing.  I'm gonna turn off that setting and undo all those changes.  That's the one thing that's hard to "undo" if we go this way.

tlambert03:
> Imma let you drive on this Talley but I'm blocking my eyes. 🙈 😝

I don't ever want @jni to have to block his eyes!  so I'm turning off the `disallow_untyped_defs` option in mypy, and I reverted all of the `) -> None:` and `) -> Any:` return annotations.  This PR is now mostly just a typing "cleanup", and an update to some of the pre-commit hook versions.

PR #42, opened 2020-07-23
Drop support for python 3.6

tlambert03:
this PR would drop support for python 3.6
(came up because I wanted to use `from __future__ import annotations` ... which requires >=3.7)


tlambert03:
Lol.  That was a resounding "yes... No discussion needed.  Merged."

Issue #34, opened 2020-07-15
Add support for Annotated type


tlambert03:
Pulling out this great tip by @VolkerH from https://github.com/napari/magicgui/issues/20#issuecomment-647888706 into a new issue.  We can use the `Annotated` type ([PEP 593](https://www.python.org/dev/peps/pep-0593/)) available through typing_extensions for parameter-specific metadata.

so, as demonstrated by @jni in https://github.com/napari/magicgui/issues/20#issuecomment-648504543, we could put [argument-specific options](https://magicgui.readthedocs.io/en/latest/configuration/#argument-specific-options) directly in the type hint:
```python

@magicgui(arg={'minimum': 10, 'maximum': 100})
def my_func(arg=50): ...

# could become this:
@magicgui
def my_func(arg: Annotated[int, {'minimum': 10, 'maximum': 100}] = 50): ...

```

tlambert03:
also want to bring this https://github.com/napari/magicgui/issues/20#issuecomment-648530931 from @jni into this thread... 

> @tlambert03: am I correct in gathering then that pretty much everything we're already doing with argument-specific and Qt-specific options could be easily added as Annotated types?

> @jni: They could, but to be honest I don't think they should. Those Qt-specific options and switching between valid widgets for a given type, and fiddling with widget rather than argument parameters should be part of the magicgui API, rather than the function signature. But something like narrowing down a type is a perfect use of Annotated imho.

@jni, what do you think?  I'm not proposing that this would be the only way to provide options in the magicgui API, just that "if an Annotated type is used, we'll parse the `__metadata__` field for options".  One additional thing I'm thinking here is that in #33, I'm kinda considering making the primary internal object that gets passed around in magicgui an [`inspect.Signature`](https://docs.python.org/3/library/inspect.html#inspect.Signature) object.  That pretty much encapsulates everything I need to build a gui (parameter name, default value, & type) ... so additional parameter-specific options on the type annotation seems natural internally, and might as well be exposed for those wanting to use it in function signatures?

VolkerH:
Good thinking using this for default slider ranges.

I made the suggestion a few minutes after I first encountered the `Annotated` type. And while I still think it is potentially a great way to address some problems in `magicgui`.
 
However (and I realize I am contradicting myself here) I also feel it is a bit odd to work with annotations to type annotations as it feels like a meta-meta level.  As an alternative to the example above, maybe it would be cleaner to have a `range` class, as in a numerical type with associated min/max values. 

[...]

PR #43, opened 2020-08-11
rewrite: proper widget protocols & signature objects

tlambert03:
**Updated 12/7/2020 after a lot of changes**

First off, I apologize for a PR that literally changes every line in the codebase 😂 ... but there was no other way!  
This PR introduces proper "interfaces" that backend apps need to implement, completely separating any Qt code (while making it much clearer what would be required of a new backend), unifying the public API, and allowing better composability of widgets/objects.  It also formalizes a close/natural connection between individual widget objects and [`inspect.Parameter`](https://docs.python.org/3/library/inspect.html#inspect.Parameter) objects; and between "container" objects (collections of widgets) and an [`inspect.Signature`](https://docs.python.org/3/library/inspect.html#inspect.Signature) objects.  ~I still need to port over some of the features (and update/write tests)~, but here's roughly how this is organized for anyone wanting to take a look:

- all interfaces are declared in [~`magicgui/bases.py`~`magicgui.widgets._protocols`](https://github.com/napari/magicgui/pull/43/files#diff-e50dfda665440ce5823f7d21ba295c157516d9e9cd7edecbe99083f09a30a137), as `typing.Protocols`, where the abstract methods that backends must implement are prefixed with `_mg_` (e.g. `_mg_get_value`).  They fall into subcategories based on their "capabilities"
   1. `WidgetProtocol`  (show/hide/enable/get_parent/get_native)
       1. `ValueWidgetProtocol`  (get/set/on_change)
           - `ButtonWidgetProtocol` (text)
           - `RangedWidgetProtocol` (min/max/step)
             - `SliderWidgetProtocol` (orientation)
           - `CategoricalWidgetProtocol` (get/set choices)
       3. `ContainerProtocol`  (list-like container for other widgets)
   2. `BaseApplicationBackend`   (similar to the vispy app)
- Backends implement those protocols using their own native widgets and stuff. I've been using composition instead of inheritance (since multiple inheritance is tricky in Qt), but either is technically fine.  See [`magicgui.backends._qtpy.widgets`](https://github.com/napari/magicgui/pull/43/files#diff-db9d4e2ba136f336bd1abc239f0d403742aa46e6ffa5f47a82c4b8d4c451c83d) for examples of backend implementations.
- classes defined in [`magicgui.widgets._bases`](https://github.com/napari/magicgui/pull/43/files#diff-0724aa4866c7759467eb2b712077951387b7428837db8bbfca5fe2bc85b87020) then wrap backend widget implementations (they expect a "widget_type") in their constructors, and define the main public api. 
- classes defined in [`magicgui.widgets._concrete`](https://github.com/napari/magicgui/pull/43/files#diff-4cd41bfe1c089eb3d610a1687d9009cb4bc4d125cafc35b67c122a14bd8458ff) (and exported at `magicgui.widgets`) are the "actual" widgets that someone might use to just directly "create a widget".  They basically just get a backend widget and use it to instantiate one of the widgets from `magicgui.widgets._bases`.  This now means that someone could create a widget directly with `magicgui.widgets.LineEdit(default='Some String')`

- the `magicgui.widgets.Container`object wraps a backend container implementation and itself implements the `MutableSequence` interface, so adding widgets to a layout just feels like appending/inserting to a list. 
   - It provides convenience classmethods (`to_signature` & `from_signature`) that convert to and from and `inspect.Signature` object.  (technically, it uses a `MagicSignature` subclass that puts some of the gui options inside the type [as an Annotated type](https://github.com/napari/magicgui/issues/34).)

- the [`magicgui.Application`](https://github.com/tlambert03/magicgui/blob/3ea582d6650e64a83b190e1dd6bc0d0e89ab5100/magicgui/application.py) object wraps the backend app, and is patterned after `vispy.use_app`.
- **The `@magicgui` decorator itself now just returns a special case of a `Container`** called a [`FunctionGui`](https://github.com/napari/magicgui/pull/43/files#diff-301d0f4470edcc7e7f8249204e7328369b7a618e0add82377ba2d72f353154e8), that wraps an `Application` and `Container` object, and binds the signature to a `__call__` method... (a nice bonus is that `inspect.signature(func)` still works just fine on a function decorated with `@magicgui`... and the decorated function is still functionally equivalent to the non-decorated function)
- mapping between types -> widgets is done in [`type_map.py`](https://github.com/napari/magicgui/pull/43/files#diff-00e91ee7ed095b8d66caa3b10aed37139a0b2602c83939a537e7acc2454d6b17).


GenevieveBuckley:
Not everything imported from `typing` is available for python versions older than 3.8. We should probably fall back to importing these from the `typing_extensions` package instead (I've made [a pull request for that on your branch here](https://github.com/tlambert03/magicgui/pull/2))

sofroniewn:
> First off, I apologize for a PR that literally changes every line in the codebase 😂 ... but there was no other way!

lol! Coming here from your comment in https://github.com/napari/napari/pull/1816#issuecomment-725682211, let me know if you'd like any help with review

tlambert03:
alright, this is ready for some poking and prodding @sofroniewn, @jni, @GenevieveBuckley (& @HagaiHargil if you're interested!).  I don't expect anyone to actually review the whole thing, but a quick read over the original post in this thread (just updated) to get the gist of the general idea of things would be appreciated.  The `examples` folder is probably a good place to go to get a sense for changes to the public API.

I've maintained the old napari examples as `OLD_napari_*` to make sure that they all still work (they just emit a lot of warnings and explain how to update to the new API).  And I'll update the readthedocs website soon too.

Some of the main changes in the new api:

- You no longer use `Gui()` to "create" the widget.  an `@magicgui` decorated function already _is_ a magicgui widget.  (it still "behaves" functionally like the wrapped function also if you just call it directly... with the one exception that "default" parameter values may be updated by the user in the GUI)
- The widget itself is no longer a `QWidget` subclass, but rather a `magicgui.widgets.Widget` subclass.  The "actual" backend widget is available at `widget.native` (and https://github.com/napari/napari/pull/1994 updates napari to look for that new object if present)
- if `func = magicgui(func)` ... then:
    - `func` is an instance of `magicgui.FunctionGui` rather than just adding the gui constructor at `func.Gui()`
    - the subwidgets themselves are now accessible at `func.<param>` rather than `func.<param>_widget`
    - changing the value of a widget is done with `func.<param>.value =` rather than `func.<param> = `
    - the "on change" signals are now at `func.<param>.changed` instead of `func.<param>_changed`
    - (all of these have passthrough deprecation warnings in the new version)
- test coverage by lines has gone down a little bit, but all of the "behavioral" tests from the old version are still in there and working.  I'll gradually bring it back up

the simplest use of magicgui is now just:
```python
from magicgui import magicgui

@magicgui
def example(x: int, y="hi"):
    return x, y

example.show(run=True)  # will take care of spinning up app if necessary
```

examples/file_dialog.py
```diff
@@ -1,11 +1,10 @@
 """FileDialog with magicgui."""
```
sofroniewn, line 1: This example isn't running for me, screen short of flashes and then code is done

examples/hard_funcs.py
```diff
@@ -1,10 +1,11 @@
 """problematic functions"""
```
sofroniewn, line 1: Here I get the following. Not sure if it is expected

```python-traceback
(base) czirwc1macos2701:magicgui nsofroniew$ python examples/hard_funcs.py
An error was encountered in the event loop:
 'functools.partial' object has no attribute '__name__'
```

examples/log_slider.py
```diff
@@ -0,0 +1,14 @@
+"""Simple demonstration of magicgui."""
```
sofroniewn, line 1: If I start moving the slider I get the following error

```python-traceback
(base) czirwc1macos2701:magicgui nsofroniew$ python examples/log_slider.py
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/widgets/_bases.py", line 336, in <lambda>
    lambda *x: self.changed(value=x[0] if x else None)
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/events.py", line 603, in __call__
    self._invoke_callback(cb, event)
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/events.py", line 625, in _invoke_callback
    cb_event=(cb, event),
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/events.py", line 619, in _invoke_callback
    cb(event)
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/widgets/_bases.py", line 812, in <lambda>
    widget.changed.connect(lambda x: self.changed(value=self))
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/events.py", line 576, in __call__
    raise RuntimeError("EventEmitter loop detected!")
RuntimeError: EventEmitter loop detected!
Abort trap: 6
```

examples/napari_param_sweep.py
```diff
@@ -26,17 +25,17 @@
     # - we contstrain the possible choices for `mode`
     @magicgui(
         auto_call=True,
-        sigma={"widget_type": QDoubleSlider, "maximum": 6, "fixedWidth": 400},
+        sigma={"widget_type": "FloatSlider", "maximum": 6},
```
sofroniewn, line 28: Very nice!!!

examples/napari_viewer.py
```diff
@@ -1,10 +1,10 @@
 """Example showing how to access the current napari viewer
```
sofroniewn, line 1: I get a future warning in this example, possible because it hasn't moved over to the new syntax yet

```python-traceback
(base) czirwc1macos2701:magicgui nsofroniew$ python examples/napari_viewer.py
/Users/nsofroniew/GitHub/magicgui/magicgui/widgets/_bases.py:658: FutureWarning:

As of magicgui 0.2.0, a `choices` callable may accept only a single positional
argument (an instance of `magicgui.widgets.CategoricalWidget`), and must return
an iterable (the choices to show). Function 'get_viewers' accepts 2 arguments.
In the future, this will raise an exception.

  FutureWarning,
/Users/nsofroniew/GitHub/magicgui/magicgui/function_gui.py:189: FutureWarning:

Creating a widget instance with `my_function.Gui()` is deprecated,
the magicgui decorator now returns a widget instance directly, so you
should simply use the function itself as a magicgui widget, or call
`my_function.show(run=True/False)` to run the application.
In a future version, the `Gui` attribute will be removed.

  FutureWarning,
```

examples/snells_law.py
```diff
@@ -16,10 +15,8 @@ class Medium(Enum):
     Air = 1.0003
 
 
-@magicgui(call_button="calculate", result={"disabled": True, "fixedWidth": 100})
-def snells_law(
-    aoi=30.0, n1=Medium.Glass, n2=Medium.Water, degrees=True,
-):
+@magicgui(call_button="calculate", result_widget=True)
```
sofroniewn, line 18: There's something funny in this example where the "result" doesn't seem to update until I shift focus away from the widget to another application, then it shows up. I'm on a mac. Can try and do more debugging / system profiling if needed. Happens with `pythonw` too.

examples/widget_demo.py
```diff
@@ -1,10 +1,10 @@
 """Widget demonstration of magicgui."""
```
sofroniewn, line 1: This widget does not have the "calculate" problem that I noted in the snells law gui case. I see output as soon as I click calculate without having to tab away.

examples/wrapt_patching.py
```diff
@@ -1,27 +1,28 @@
 """Utilities for changing function signatures."""
```
sofroniewn, line 1: I currently get an error with this example

```python-traceback
(base) czirwc1macos2701:magicgui nsofroniew$ python examples/wrapt_patching.py
  File "examples/wrapt_patching.py", line 305, in <module>
    guis[func.__name__] = magicgui(func, **opts)
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/function_gui.py", line 283, in magicgui
    return inner_func(function)
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/function_gui.py", line 275, in inner_func
    app=app,
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/function_gui.py", line 74, in __init__
    sig = magic_signature(function, gui_options=param_options)
  File "/Users/nsofroniew/GitHub/magicgui/magicgui/signature.py", line 270, in magic_signature
    "keyword arguments (gui_options) MUST match parameters in the "
ValueError: keyword arguments (gui_options) MUST match parameters in the decorated function.
Got extra keys: {'ignore'}
```

sofroniewn (commented):
This is looking great @tlambert03. I went through all the examples, found a couple bugs / unexpected things, but overall very exciting. I'll have a look over more of the code tomorrow.

tlambert03:
> This is looking great @tlambert03. I went through all the examples, found a couple bugs / unexpected things, but overall very exciting. I'll have a look over more of the code tomorrow.

thanks so much!  `hard_funcs` and `wrapt_patching` shouldn't be in examples yet.  Those were experiments... I'll remove them until they're working.  I'll go through and fix all the others though!

examples/hard_funcs.py
tlambert03: will remove from examples for now

examples/napari_viewer.py
tlambert03: going to remove this altogether since it's confusing to know whether this is necessary or not (this sort of napari type lookup will be in `napari.utils._magicgui`)

examples/snells_law.py
tlambert03: hmmm, I can't repeat this... maybe be something different between our envs/OS?

examples/widget_demo.py
tlambert03: very interesting!  but also mysterious! :)  i'm stumped (but... still can't repeat the phenotype of the snells_law problem)

examples/wrapt_patching.py
tlambert03: removing

examples/file_dialog.py
tlambert03: fixed

HagaiHargil:
Congrats! I'll try looking at it tomorrow if it will still be relevant, perhaps I'll stumble upon the same issues as @sofroniewn has.

What is the `run` keyword for the `show` method? Does that replace the `event_loop` context manager?

tlambert03:
> Congrats! I'll try looking at it tomorrow if it will still be relevant, perhaps I'll stumble upon the same issues as @sofroniewn has.

thanks! no pressure.  will still be relevant tomorrow.  hopefully some of those issues are gone now.

> What is the `run` keyword for the `show` method? Does that replace the `event_loop` context manager?

yeah, `function.show(run=True)` is a shortcut for starting the event loop.  It will block immediately and show the gui.  The `event_loop` context manager is still there though, in case you wanted to setup multiple independent widgets or something before blocking (as in the `file_dialog` example):
```python
with event_loop():
    widget_a.show()
    widget_b.show()
```

examples/log_slider.py
tlambert03: hmm. this is fixed by commenting out the `if self._emitting: raise RuntimeError` in events.py... and nothing else seems to break, but I need to investigate a bit more

GenevieveBuckley:
FYI I'm sorry, but I don't have the bandwidth to take a look at this right now

examples/snells_law.py
sofroniewn: yeah, i'm not so concerned about this, but funny that I don't see it in the widget_demo too ...

magicgui/function_gui.py
```diff
@@ -0,0 +1,286 @@
...
+    function: Optional[F] = None,
+    *,
+    orientation: str = "horizontal",
+    labels: bool = True,
+    call_button: Union[bool, str] = False,
+    auto_call: bool = False,
+    result_widget: bool = False,
+    app: AppRef = None,
+    **param_options: dict,
+):
```
sofroniewn, line 218: If I'm not mistaken, this is really the core API for magicgui, which we're planning to allow users to access in https://github.com/napari/napari/pull/1856 via the `magic_kargs` argument?

magicgui/function_gui.py
```diff
@@ -0,0 +1,286 @@
...
+        The type of layout to use. Must be one of {'horizontal', 'vertical',
+        'form', 'grid'}, by default "horizontal"
+    labels : bool
+        Whether labels are shown in the widget. by default True
+    call_button : bool or str, optional
+        If True, create an additional button that calls the original function when
+        clicked.  If a ``str``, set the button text. by default False
+    auto_call : bool, optional
+        If True, changing any parameter in either the GUI or the widget attributes
+        will call the original function with the current settings. by default False
```
sofroniewn, line 236: Looks like `app` doc string is missing (not trying to be pedantic, but just flagging in case accidentally missed)

magicgui/function_gui.py
```diff
@@ -0,0 +1,286 @@
...
+
+    **param_options : dict of dict
+        Any additional keyword arguments will be used as parameter-specific options.
+        Keywords MUST match the name of one of the arguments in the function
+        signature, and the value MUST be a dict.
```
sofroniewn, line 241: Am I correct in thinking that the values dicts must line up for with the acceptable widget types. Ultimately we'll want to link out here to the available widgets etc. (I know it's early to be thinking about that, and that can come later, just making sure I'm parsing everything critical right).

magicgui/widgets/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    "LogSlider",
+    "PushButton",
+    "RadioButton",
+    "RangeEdit",
+    "SliceEdit",
+    "Slider",
+    "SpinBox",
+    "TextEdit",
+    "Widget",
+]
```
sofroniewn, line 54: Am I right in thinking this is the core list of widgets we support and their names?

I'm taking a look now at other control widgets provided in other ecosystems like `Shiny` https://shiny.rstudio.com/tutorial/written-tutorial/lesson3/ and `ipywidgets` https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html and it looks like we have good coverage with our list.

Do you think there are benefits that could come from standardizing our API with the ipywidgets case, both at the level of the widgets we provide and their input parameters?

With the new 0.2.0 architecture could one in fact one day imagine an HTML backend supported by ipywidgets, no? Or alternatively if we could match APIs, could we think about some of this effort as now providing a Qt backend to the ipywidgets interface along with magic to go from a function definition to that widgets interface?

Note that I'm absolutely fine to merge this PR as is without change, but curious how you think about this, and maybe this could be a direction the library could move.

sofroniewn (commented):
Thanks for those fixes @tlambert03! The only one still not working for me is the log slider, which sounds like your looking into.

I looked through some of the code / core functionality, it's looking great. I left a couple comments/ questions that mostly represent my musings and thoughts as I read through things. Probably the most interesting is how you see the API of our widgets relating to the API of [ipywidgets](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html). As I said in my comment no need to change anything in this PR, but curious of your thoughts, especially if one day napari might want to support an html frontend!

magicgui/type_map.py
```diff
@@ -0,0 +1,272 @@
...
+        A widget class from the current backend that should be used whenever ``type_``
+        is used as the type annotation for an argument in a decorated function,
+        by default None
+    return_callback: callable, optional
+        If provided, whenever ``type_`` is declared as the return type of a decorated
+        function, ``return_callback(widget, value, return_type)`` will be called
+        whenever the decorated function is called... where ``widget`` is the MagicGui
+        instance, and ``value`` is the return value of the decorated function.
+
+    OPTIONS
```
jni, line 213: ...? Why are you yelling at me? 😭 =P

jni (approved):
Thanks @tlambert03! I've only skimmed the main code, looked at the examples and it's a great change from a usage perspective. Two things:

- I agree with @sofroniewn's proposal to aim to unify our widget names with other frameworks in the ecosystem, ipywidgets being a very popular one. Honestly I haven't looked around to see if it's *the* most popular one. But, small changes now (Slider -> IntSlider) might make a big difference when it comes to adding an ipywidgets backend.
- aesthetically, I don't love the `_mg_` prefix on all the methods to make a backend. I get the motivation, but it doesn't feel like the best choice. I think the worst of it is in my head, `_mg_` means milligrams, or magnesium, or the car manufacturer, but certainly not magicgui. Maybe that will change with time, the way Jupiter with an i is weird now. =P But maybe `_mgui_` would be better, even if longer?

At any rate, awesome work as always.

tlambert03:
@jni @sofroniewn: Thank you both!  I know it takes a lot of mental effort to even open this issue, let alone look at the code.  So I really appreciate your time.  I agree on all accounts: ipywidgets seems like a good API to mimic (and I'll do a little more research to see if there are other strong contenders in the python ecosystem).  I was actually just yesterday playing a bit with ipywidgets to see how hard it would be to create a backend for it.  The `magicgui` decorator is in fact quite similar to the [`ipywidgets.interact` function](https://ipywidgets.readthedocs.io/en/latest/examples/Using%20Interact.html), so I'm _hoping_ that it should be very doable.

also agree on changing `_mg_` to `_mgui_` (and, can go farther if you don't like the prefix idea at all?  That was kinda inspired by looking at [vispy](https://github.com/vispy/vispy/blob/a6ec64bb7b56739b463e3c16245a31940ca4f5f7/vispy/app/backends/_template.py))

magicgui/function_gui.py
```diff
@@ -0,0 +1,286 @@
...
+from magicgui.application import Application, AppRef
+from magicgui.events import EventEmitter
+from magicgui.signature import magic_signature
+from magicgui.type_map import _type2callback
+from magicgui.widgets import Container, LineEdit, PushButton
+from magicgui.widgets._protocols import ContainerProtocol
+
+
+class FunctionGui(Container):
+    """Wrapper for a container of widgets representing a callable object.
```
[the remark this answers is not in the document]
tlambert03: hmmm, I guess short of wanting to subclass and modify `FunctionGui`, probably much more likely that they just use the `magicgui` function.  Any thoughts on that?

examples/callable.py
```diff
@@ -0,0 +1,24 @@
...
+    return f"{y} {x}"
+
+
+def g(x: int = 6, y="another string") -> str:
+    return f"{y} asdfsdf {x}"
+
+
+@magicgui(call_button=True, func={"choices": ["f", "g"]})
+def example(func="f"):
+    pass
```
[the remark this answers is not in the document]
tlambert03: yeah that seems nice.  lemme look into it

examples/file_dialog.py
```diff
@@ -1,11 +1,10 @@
...
 from pathlib import Path
-from magicgui import event_loop, magicgui
 from typing import Sequence
 
+from magicgui import event_loop, magicgui
+
 
-# may also add Qt-style filter to filename options:
-# e.g. {"filter": "Images (*.tif *.tiff)"}
```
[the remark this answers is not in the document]
tlambert03: right, forgot to get back to that.  Yes, to be implemented.  Wanted to think of the best non-qt specific approach.  but I guess a generic string passed to the backend is a decent place to start.

magicgui/function_gui.py
tlambert03: good catch! thanks

magicgui/function_gui.py
tlambert03: Yeah, good intuition.  You're exactly right.  Not all key/values will work for all widget types.  I [started](https://github.com/napari/magicgui/blob/8965c9402ddb74b323633dd01de30453e5072ff2/magicgui/types.py#L46) on creating some typing support for that, widget-specific options still needs work/clarification.

magicgui/widgets/__init__.py
tlambert03: > Do you think there are benefits that could come from standardizing our API with the ipywidgets case, both at the level of the widgets we provide and their input parameters?

yes, and now would be a good time to do it.  Definitely might help to keep the naming consistent... not immediately sure about the parameters they each accept, but if it works out well, why not!

> With the new 0.2.0 architecture could one in fact one day imagine an HTML backend supported by ipywidgets, no? 

def.  as mentioned elsewhere, I was already starting to look into this.  i think it should be doable (if not a bit overlapping in functionality with the `ipywidgets.interact` function).

> Note that I'm absolutely fine to merge this PR as is without change, but curious how you think about this, and maybe this could be a direction the library could move.

I think we should rename widgets here as desired, and then tackle widget options later.

examples/snells_law.py
HagaiHargil: I can't reproduce it as well (Linux).

magicgui/function_gui.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Parameters
+    ----------
+    function : Callable, optional
+        The function to decorate.  Optional to allow bare decorator with optional
+        arguments. by default None
+    orientation : str, optional
+        The type of layout to use. Must be one of {'horizontal', 'vertical',
+        'form', 'grid'}, by default "horizontal"
+    labels : bool
```
HagaiHargil, line 229: Should say "bool, optional"

examples/snells_law.py
tlambert03: thanks for checking!

magicgui/function_gui.py
```diff
@@ -0,0 +1,291 @@
...
+    **param_options : dict of dict
+        Any additional keyword arguments will be used as parameter-specific options.
+        Keywords MUST match the name of one of the arguments in the function
+        signature, and the value MUST be a dict.
+
+    Returns
+    -------
+    Callable
+        The original function is returned with a new attribute ``Gui``.  Gui is a
+        subclass of MagicGui that, when instantiated, will create a widget representing
```
HagaiHargil, line 252: I see a few mentions of MagicGui in the codebase. Is that FunctionGui? Or do you refer to the actual package?

magicgui/function_gui.py
```diff
@@ -0,0 +1,291 @@
...
+        the signature of the original function.  Furthermore, *calling* that widget will
+        call the original function using the state of the Gui arguments.
+
+    Examples
+    --------
+    >>> @magicgui
+    ... def my_function(a: int = 1, b: str = 'hello'):
+    ...     pass
+    ...
+    ... gui = my_function.Gui(show=True)
```
HagaiHargil, line 262: Is that correct? Shouldn't be updates to `.show(run=True)`?

magicgui/function_gui.py
tlambert03: thank you!  that's vestigial...  will update.

magicgui/function_gui.py
tlambert03: same as above.  looks like I need to do some find/replace in my docstrings :)  thanks for the sharp eyes

HagaiHargil:
Looks great, well done! I really think you should sell this package to a broader audience - I use it all the time for my work, independently of `napari`. I believe it elegantly solves an issue many data scientists face on a daily basis.

magicgui/function_gui.py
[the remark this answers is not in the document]
sofroniewn: yeah, that makes sense!!

magicgui/widgets/__init__.py
sofroniewn: > I think we should rename widgets here as desired, and then tackle widget options later.

This sounds like a reasonable step (but not required). I'll post a longer response in the main thread, very exciting though!!

sofroniewn (approved):
>  I agree on all accounts: ipywidgets seems like a good API to mimic (and I'll do a little more research to see if there are other strong contenders in the python ecosystem). I was actually just yesterday playing a bit with ipywidgets to see how hard it would be to create a backend for it. The magicgui decorator is in fact quite similar to the ipywidgets.interact function, so I'm hoping that it should be very doable.

That's great! I see `ipywidgets` here as analogous to `qtpy` and that having a magicgui backend for it would be incredibly powerful. Given that I don't think we need to absolutely conform to the `ipywidgets` API in the same way that we don't conform to the `qtpy` API, so I would be fine with `Slider` instead of `IntSlider` if we wanted, but I'm just glad we're thinking about it and the `ipywidgets.interact` function too, which you're right is highly related.

One could imagine other backends too and even moving towards an intermediate json-schema like representation if we wanted. Today someone else pointed me towards https://github.com/rjsf-team/react-jsonschema-form, which is a different use case, but a related idea. I'm also reminded of the @freeman-lab creation https://github.com/freeman-lab/control-panel from a while back!

All this to say, I really like the new direction. I think it opens up lots of possibility for magicGui to be used well beyond napari too!

I'm approving now too. We should probably fix the log slider example before merge, but everything else I'm good with!

tlambert03:
> Looks great, well done! I really think you should sell this package to a broader audience - I use it all the time for my work, independently of napari. I believe it elegantly solves an issue many data scientists face on a daily basis.

Thanks, and thanks a lot for taking the time to have a look.  If you have any non-napari examples that you'd consider contributing to the examples, I'd happily include them in the new docs i'm working on  (it's based on jupyter book and show be able to show the widget result inline for a given bit of code)

examples/callable.py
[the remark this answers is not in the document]
tlambert03: going to make a new issue to look into this later

examples/file_dialog.py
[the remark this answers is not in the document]
tlambert03: returned

examples/log_slider.py
tlambert03: fixed (this was caused by the result widget emitting a signal as the `auto_call` functionality was updating it)

tlambert03:
> I don't think we need to absolutely conform to the ipywidgets API in the same way that we don't conform to the qtpy API, so I would be fine with Slider instead of IntSlider if we wanted, but I'm just glad we're thinking about it and the ipywidgets.interact function too

great. maybe I'll address that it a followup PR (but before this actually gets released), just to simplify the diffs moving forward.

> All this to say, I really like the new direction. I think it opens up lots of possibility for magicGui to be used well beyond napari too!

Glad you like it.  I hope it does open up some avenues... just starting to chat with @oeway as well about possibly making a backend that might work for imjoy too

> We should probably fix the log slider example before merge, but everything else I'm good with!

cool!  There's always more to be done, so I'm thinking I'll take one last look through, and then merge so the changes can go back to being incremental.  If someone were to check out master at this point it _should_ be very useable.

Thanks again everyone for dealing with a horrifically large PR.

Issue #45, opened 2020-11-11
Widget label editable 


sfcaracciolo:
I create this feature request like @tlambert03 have suggested here https://forum.image.sc/t/magicgui-and-docked-widget-questions/44983 

It would be fine, add a "display_name" parameter on the @magicgui decorator to personalize the string showed left to the widget.

For instance, this fragment:

```python
import napari
from magicgui import magicgui

@magicgui()
def test_gui(var_name: int = 0):
    pass

with napari.gui_qt():
    widget = test_gui.Gui(show=True)
```

it has this output, where "var_name" label is not editable

![Screenshot_2020-11-11_19-10-21](https://user-images.githubusercontent.com/41028670/98870103-91ae4400-2451-11eb-9ddd-90c7c89d420a.png)

@jni suggested using a "display_name" option on the @magicgui decorator like this:

```python
@magicgui(var_name={"display_name": "some editable string"})
def test_gui(var_name: int = 0):
    pass
```

Cheers!

tlambert03:
implemented in #48, and carried over into the new API coming in v0.2.0

PR #48, opened 2020-11-22
Add a "display_name" option to modify a widget's label (#45)

HagaiHargil:
Addresses #45.

This PR adds the option to modify a widget's label by including a `display_name` key in that argument's specific dictionary.

It also adds a paragraph in the docs, does a small change to the examples section that shows this behavior and adds a test.

### DIFF
--- a/magicgui/_tests/test_magicgui.py
+++ b/magicgui/_tests/test_magicgui.py
@@ -66,11 +66,10 @@ def test_overriding_widget_type(qtbot):
     assert gui.a == "1"
 
 
-def test_overriding_arg_display_name(qtbot):
-    """Test that a new label is given to a widget based on
-    the 'display_name' key."""
+def test_overriding_arg_label(qtbot):
+    """Test that `label` overrides the default label text."""
 
-    @magicgui(a={"display_name": "b"})
+    @magicgui(a={"label": "b"})
     def func(a: int = 1):
         pass
 
--- a/magicgui/_tests/test_magicgui.py
+++ b/magicgui/_tests/test_magicgui.py
@@ -66,6 +66,18 @@ def test_overriding_widget_type(qtbot):
     assert gui.a == "1"
 
 
+def test_overriding_arg_display_name(qtbot):
+    """Test that a new label is given to a widget based on
+    the 'display_name' key."""
+
+    @magicgui(a={"display_name": "b"})
+    def func(a: int = 1):
+        pass
+
+    gui = func.Gui()
+    assert gui.children()[-1].text() == "b"
+
+
 def test_overriding_arg_type(qtbot):
     """Test overriding the widget type of a parameter."""
     # a will now be a LineEdit instead of a spinbox
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -47,8 +47,8 @@ def my_func(arg=5):
 ### `display_name`
 
 By default, the label which will be displayed next to the widget will have the
-variable's name. If you wish to modify that you may include in the argument's
-options dictionary the `display_name` key with the desired label:
+variable's name. If you wish to modify that, add a `display_name` entry to the argument's
+options dictionary with the desired label:
 
 ```python
 @magicgui(user_address={'display_name': 'Address'})
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,14 +44,14 @@ def my_func(arg=5):
     ...
 ```
 
-### `display_name`
+### `label`
 
 By default, the label which will be displayed next to the widget will have the
-variable's name. If you wish to modify that, add a `display_name` entry to the argument's
+variable's name. If you wish to modify that, add a `label` entry to the argument's
 options dictionary with the desired label:
 
 ```python
-@magicgui(user_address={'display_name': 'Address'})
+@magicgui(user_address={'label': 'Address'})
 def get_address(user_address: str):
     ...
 ```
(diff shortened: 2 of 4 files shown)


tlambert03:
thank you for implementing this @HagaiHargil!  and thanks for such a thorough treatment (docs, examples, etc...)!
My only question (and this is for @jni as much as you, since he originally proposed the "display_name" paramater), is whether we might want to name it "label" instead?  I can imagine using the same key for `label=False` to hide a specific label (down the road, when we add that sort of granularity to the labels).  so if it's a bool, the label is shown or hidden as indicated, and if it's a string, it's used as the display name.  What do you both think?

GenevieveBuckley:
> My only question (and this is for @jni as much as you, since he originally proposed the "display_name" paramater), is whether we might want to name it "label" instead? I can imagine using the same key for `label=False` to hide a specific label (down the road, when we add that sort of granularity to the labels). so if it's a bool, the label is shown or hidden as indicated, and if it's a string, it's used as the display name. What do you both think?

I feel like:
* "label" is much more ambiguous than "display_name" (I would probably confuse "label" with the label image layer or something like it if I saw that)
* I find the idea of cramming multiple functionality into a single kwarg (or dictionary key, in this case) potentially confusing. Would it be simpler to add another dictionary key called something like "hide_display_name" with a boolean value, if you wanted to add this kind of functionality later on?

tlambert03:
> (I would probably confuse "label" with the label image layer or something like it if I saw that)

i don't feel like there should be any implicit/assumed relationship between magicgui and napari... so that particular concern doesn't resonate with me that much.  The word next to the widget is a literal `QLabel` and "label widgets" are usually read-only text strings, so i feel like "label" is the natural thing to call it here.

> I find the idea of cramming multiple functionality into a single kwarg (or dictionary key, in this case) potentially confusing

that I can understand.  I _kinda_ liked just saying `label=False` or `label='something different'` ... but `label='something different' and `hide_label=True` is totally fine with me too.

HagaiHargil:
I'm 100% behind @tlambert03 on this one. I thought "label" was the better option but went with "display_name" due to it being mentioned in the original issue. I'll happily replace all use cases of one with the other.

My use cases for magicgui are completely independent from napari, so I don't think that there would be any confusion with napari's Label keyword, especially since we're talking about a key within a dictionary within a decorator - all of them are exclusively magicgui's. Moreover, in many common data science APIs there are keywords which behave like @tlambert03 proposed - `False` or `None` if you don't want it around, and some value otherwise.

GenevieveBuckley:
> I'm 100% behind @tlambert03 on this one. I thought "label" was the better option but went with "display_name" due to it being mentioned in the original issue. I'll happily replace all use cases of one with the other.

I'm happy to be outvoted. Thanks for making this @HagaiHargil

tlambert03:
> > I'm 100% behind @tlambert03 on this one. I thought "label" was the better option but went with "display_name" due to it being mentioned in the original issue. I'll happily replace all use cases of one with the other.
> 
> 
> 
> I'm happy to be outvoted. Thanks for making this @HagaiHargil 
> 
> 
> 
> 

@jni, care to weigh in on this and put us back in gridlock? 😂

magicgui/_tests/test_magicgui.py
```diff
@@ -66,6 +66,18 @@ def func(a: int = 1):
     assert gui.a == "1"
 
 
+def test_overriding_arg_display_name(qtbot):
+    """Test that a new label is given to a widget based on
+    the 'display_name' key."""
```
tlambert03, line 71:
```suggestion
    """Test that `display_name` overrides the label text."""
```

(fixes [linting error](https://github.com/napari/magicgui/pull/48/checks?check_run_id=1438037299#step:5:14))

jni:
> @jni, care to weigh in on this and put us back in gridlock? 😂

My reading is that "label" is itself kinda Qt-specific. I would find it confusing. How about just "name"?

jni:
I agree with @GenevieveBuckley about separating the APIs, too. So, `name='my widget', display_name=True`? I might also suggest that display_name defaults to True if name is given, False otherwise.

tlambert03:
> > @jni, care to weigh in on this and put us back in gridlock? 😂
> 
> 
> 
> My reading is that "label" is itself kinda Qt-specific. I would find it confusing. How about just "name"?

I don't think it is... tkinter also uses it, and Wikipedia uses it: https://en.m.wikipedia.org/wiki/Label_(control)

jni:
Ok fine 😂 I knew as I was writing it that I would regret it. I'll rephrase: I would be confused by the word "label", and I think other people who are not super familiar with GUI frameworks, ie *exactly the target audience for magicgui*, might also be confused!

I don't feel suuuuper strongly about this, but "name" would easily be my preference.

tlambert03:
> I think other people who are not super familiar with GUI frameworks, ie *exactly the target audience for magicgui*, might also be confused!
> 
> 
> I don't feel suuuuper strongly about this, but "name" would easily be my preference.

I think "name" is also kinda vague... and I don't think "label" is super jargony. When you want to put a sticker on something to say "what is this thing"... you call it a label.  I think the magicgui target audience will figure it out fine.  I'd like to stick with la el

tlambert03:
Thanks @HagaiHargil, we can leave the discussion of hide/show individual labels for later, when we actually support that sort of thing (which probably should wait for #43 anyway).
I appreciate this submission!

tlambert03 (approved):

HagaiHargil:
Happy to introduce "label".

tlambert03 (approved):

PR #61, opened 2020-12-24
Add Table Widget

tlambert03:
Still working on it, but mostly done... this adds a Table widget that would work for a pandas dataframe, or anything that pandas can turn into a dataframe (like a 2D numpy array, a list of dicts, a dict of lists, etc...)

```python
In [1]: from magicgui.widgets import Table

In [2]: data = {'col_1': [3, 2, 1, 0], 'col_2': ['a', 'b', 'c', 'd']}

In [3]: t = Table(value=data)

In [4]: t.show()
Out[4]:
Table(name={self.name})
   col_1 col_2
0      3     a
1      2     b
2      1     c
3      0     d
```

Screen Shot 2020-12-23 at 10 38 52 PM


tlambert03:
haven't decided yet if i'm actually going to require pandas to use tables, or use it only if the input data is already a dataframe...  input welcome

tlambert03:
This is working pretty well for a start!  I'm happy to merge at this point, and work on hooking it up to `type_map` in the future (i.e. nothing will automatically create a table widget yet)

magicgui/widgets/_table.py
```diff
@@ -0,0 +1,134 @@
...
+        else:
+            data = list(zip(*data.values()))
+
+    else:  # pragma: no cover
+        raise ValueError("only recognize index or columns for orient")
+    return data, index, columns or ()
+
+
+def _value_to_table_data(value: Any) -> Tuple[list, tuple, tuple]:
+    """Convert arbitrary value into appropriate table data format."""
```
jni, line 94: imho this docstring should be expanded. `value` is hardly "arbitrary", but rather one of several formats, which I don't see documented anywhere. Is dict-of-dicts common? I have never seen it before and I don't love it. And I haven't seen the 3-tuple of (data, index, columns), either... Do we not want to use the dict-of-columns as the default format, which would then make it directly compatible with napari's properties tables? (Not to mention pandas?)

jni (commented):
@tlambert03 I have some questions about the API. Since this is a new feature I would suggest leaving this until 0.2.1, since it is not a breaking change to add it, but it *is* a breaking change to change the values etc.

If you're convinced this is the right format though (I've only thought about this for 10 min 😅), feel free to go ahead! My objection is a weak one.

magicgui/widgets/_table.py
tlambert03: yeah, def... will expand the docstring.
All these formats are things that I found when digging through the formats that a DataFrame takes.  The dict-of-dicts is what you get if you do [`DataFrame.to_dict()`](https://github.com/pandas-dev/pandas/blob/aad85ad1e39568f87cdae38fece41cd2370234a7/pandas/core/frame.py#L1496-L1497)  (i looked through their API for the most "pure python type" version of a dataframe, and that's what I found...  it does seem like the only way to get data, headers, and index headers all in one, so.. maybe people use it?)
As for the default format... (a 2d data array, a list of headers and a list of row indexes), that was borrowed from the [internal Dataframe representation](https://github.com/pandas-dev/pandas/blob/aad85ad1e39568f87cdae38fece41cd2370234a7/pandas/core/frame.py#L506-L508), which has a 2D array of `.values`, a 1D `.index` and a 1D `.columns`.  so compatibility with pandas was pretty much my first priority here.  That said, I'm happy to have some different representation be the default when you do `table.value`!  thoughts?

tlambert03: I should add, I'm more than happy with a dict of columns... but that excludes row headers (should someone want them)

tlambert03: (also, this can all wait 😄 ... will do 0.2.0 without and we can discuss later)

jni: > internal Dataframe representation

Well, I think that's glossing over some rather major subtleties because of dtypes — see this rather excellent blog post from Uwe Korn:

https://uwekorn.com/2020/05/24/the-one-pandas-internal.html

imho row headers are less useful and accessible via `reset_index` anyway...

tlambert03: sorry, I wasn't suggesting that pandas _actually_ maintains something like a numpy array and two lists internally... but it does take it as an input, and if you do `df.values`, `df.index`, `df.columns`, that's what you get.
as mentioned, I'm happy to use a dict of column arrays.  but even if row headers are less useful, there should be a way for someone to provide them and query them here (outside of using pandas), so how do you want that to look?

tlambert03: @jni ... don't want this to get too far off the radar.  It's mostly waiting on your approval of the API at this point.  I don't care too much what the data structure is but I wanted it to be:
- pure python
- easy to convert to/from a dataframe
- able to represent table data, row headers, and column headers

beyond that, whatever you want is fine with me

tlambert03:
I created a new `TableData` object.  It mimics the pandas dataframe API (in the sense that it has `.values`, `.index`, and `.columns` attributes that return the same format of data, and the constructor is similar). The user can set data with a variety of formats.  But the data returned from the widget will, for now, always be an instance of `TableData`, with those three attributes.

sofroniewn (commented):
Just had a quick read though and it looks good, but I can play with some toy examples tomorrow to see how if feels and see if I can annotate some functions with inputs/ outputs that are tables

magicgui/widgets/_table.py
jni: @tlambert03 

> even if row headers are less useful, there should be a way for someone to provide them and query them here (outside of using pandas),

I'd aim to mirror the `read_csv` API, with the index included as a column and `index_col='column_name'` as part of the API. How does that sound?

jni:
> I created a new TableData object. It mimics the pandas dataframe API (in the sense that it has .values, .index, and .columns attributes that return the same format of data, and the constructor is similar)

I really would prefer to avoid a new object, and suggest either just biting the bullet and using pandas dataframes, using standard Python types (discussed above), or waiting/participating in discussions with the [data-APIs](https://data-apis.org/) folks (currently there's not even a dataframe draft standard, though they are working on it) and [Apache Arrow](https://arrow.apache.org/).

magicgui/widgets/_table.py
tlambert03: > I'd aim to mirror the read_csv API, with the index included as a column and index_col='column_name' as part of the API. How does that sound?

I _think_ I'm ok with that, but i want to just make sure we're clear on what part of the API we're discussing here. What this really comes down to (at the moment) is not what the user has to provide to magicgui (They're going to be able to set the value with a lot of standard formats... when setting `widget.value` ) it's what object the user gets _back_ when retrieving the value from the widget.  So are you proposing this?

```python
# one of many valid input formats
widget = Table()
widget.value = [[1,2,3],[4,5,6]]
# get the value back
widget.value

# returns
{0: [1,4], 1: [2,5], 2: [3,6], 'index': [0,1], 'index_col': 'index'}
```

something tells me that's _not_ what you're imagining (I'm not a fan myself).  I'm quite happy to return a tuple as I started with `widget.value -> ([[1,2,3],[4,5,6]], [], [])`, or a `NamedTuple`...  I really don't want to require someone to depend on pandas just to provide a simple 2x2 table with text.  I feel rather firm on that for this stage of development.  I made the `TableData` object to have something like a named tuple that would provide `data.value`, `data.index`, `data.columns` so that if we _do_ go to a dataframe in the future, the user just gets _more_ API, (and the existing API doesn't break).

If you don't like any of the dict return shown above, the tuple return value, or the `TableData` object, please provide an example of the pure python object that you want the user to get back when they query the current value of the widget.

tlambert03: It occurs to me that you might look at the example above and say "if they give me `widget.value = [[1,2,3],[4,5,6]]` then that's what they should get back".  So please also imagine that they provided a more complex object in, like one of the supported dict formats that also provides header info

tlambert03: another option is that we extend the `ValueWidget` API just for `Table` widgets, so that they can get and set `widget.index` and `widget.columns`, reducing `widget.value` to a simple 2x2 array that never has header information

tlambert03: ⬆️ I think this is the best approach... and will work on that unless you object

jni: On phone, but briefly, imho it would be great to have .values (or .data, or .properties, I'm agnostic here and I do think it might be good to differentiate from a common API if we don't conform to it) be a dict-of-columns, and *optionally* a string attribute index_col that lets the user know which is the "special" column, if any, that should be used as index. 

Does that meet your requirements for everything that needs to be represented?

tlambert03:
okay!  we have a much improved (I think) widget, with a dict-like primary API, and a numpy like `DataView` attribute at `widget.data`.  

see [`examples/table.py`](https://github.com/tlambert03/magicgui/blob/add-table-widget/examples/table.py) for a quick rundown of the API, or look at the tests.

thank you @jni for the feedback and guidance

Attempts were made to minimize cell get/set access as much as possible when looping, but making this very performant with large datasets is going to be a long-term goal.

tlambert03:
ok, I think this PR is ready and tests are up at 97% coverage.  @jni I added support for "joins" in data with non-matching rows/columns... (for now that logic requires pandas and will raise a `ValueError` otherwise).

I'll leave this open for a few days, if anyone wants more time to look, that's fine too, just say "please wait... want more time"

sofroniewn:
> I'll leave this open for a few days, if anyone wants more time to look, that's fine too, just say "please wait... want more time"

I can try and give this another spin tomorrow!

PR #66, opened 2020-12-27
resolve ForwardRef on widget.annotation

### LINKED ISSUE #65: Should we permanently resolve ForwardRef type annotations?
In https://github.com/napari/napari/pull/1856#discussion_r549041360, @sofroniewn shows an issue in which napari's callback function fails when a ForwardRef is used as a type annotation.  While #52 takes care of this internally... currently, it's up to all dependent libraries to do it for themselves.  It would probably be easy enough to make sure that `widget.annotation` is always a resolved ForwardRef?

tlambert03:
closes #65 
I think this is probably the right thing to do... We will have already tried to resolve any forward ref annotations when creating widgets, so it should already be in `sys.modules` and we might as well store the resolved type on `widget.annotation` for the convenience of others.

### DIFF
--- a/magicgui/widgets/_bases.py
+++ b/magicgui/widgets/_bases.py
@@ -53,6 +53,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    ForwardRef,
     List,
     MutableSequence,
     Optional,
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
Just verified this works as expected for my napari usecases in https://github.com/napari/napari/pull/1856

PR #69, opened 2020-12-27
Fix parent_changed signal emission

tlambert03:
This PR cleans up some of the `parent.changed` signal emission logic, and improves the event filter used to detect when a Qt Widget has had its parent changed


sofroniewn (approved):
LGTM! Much nicer not to have to use the `QTimer`

PR #73, opened 2020-12-29
Resolve ForwardRefs on return annotations

tlambert03:
This PR also resolves ForwardRefs on return type annotations (#66 only handled ForwardRefs on parameter annotations).

I also changed it so that `register_type` immediately resolves any ForwardRefs.  It's unlikely that someone would be passing a forwardref there, but it will also work now.


sofroniewn (approved):
Just confirmed this now works as expected and fixes https://github.com/napari/napari/issues/2028#issuecomment-752183003. Thanks!!

PR #82, opened 2021-01-05
Convert maximum/minimum kwargs to max/min and warn

tlambert03:
One deprecation warning slipped through after #60 ... for those using `maximum` and `minimum` kwargs on their widgets, this will catch those kwargs and turn them to max/min and show a warning

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,14 @@
+ignore:
+  - magicgui/_version.py
+coverage:
+  status:
+    project:
+      default:
+        target: auto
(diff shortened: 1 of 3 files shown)


sofroniewn (approved):
LGTM! One could consider aliasing to allow both, but not that big a deal either way. Probably easier just to stick to one for now

tlambert03:
I think in the medium term, what I might like to do is let backends take a crack at parsing kwargs and returning them into a standardized magicgui form.  That might let us "seem to" support both ipywidgets and qt, provided someone has done `use_app('qt')`.  but yeah, something for later

Issue #81, opened 2021-01-05
magicgui.widgets.CategoricalWidget not found in magicgui 0.2.1


haesleinhuepf:
**Describe the bug**
Hi Talley @tlambert03,

tl;dr: Is there example code available for using choices in magicgui 0.2.1? I get error messages while attempting to use `CategoricalWidget`.

I'm just updating the magicgui dependency in [my experimental napari plugin](https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/master/napari_pyclesperanto_assistant/_operations/_operations.py#L12) and receive this error message in the context of having a `choice`:

```
As of magicgui 0.2.0, a `choices` callable may accept only a single positional
argument (an instance of `magicgui.widgets.CategoricalWidget`), and must return
an iterable (the choices to show). Function 'get_layers' accepts 2 arguments.
In the future, this will raise an exception.
```

However, the mentioned class `magicgui.widgets.CategoricalWidget` does not exist. If I import it, I receive this error message:

```
(clustering) C:\structure\code\napari_pyclesperanto_assistant\docs\demo>python start_assistant.py
C:\Users\rober\miniconda3\envs\clustering\lib\site-packages\magicgui\_qt\widgets.py:13: FutureWarning:

The 'magicgui._qt.widgets' module has been removed.
For 'QDataComboBox', please use 'magicgui.widgets.ComboBox' directly.
(Or just use the string `widget_type='ComboBox'`)
In the future this will raise an exception.

  warnings.warn(
I'm asked:  C:\structure\code\napari_pyclesperanto_assistant\napari_pyclesperanto_assistant\data\Lund_000500_resampled-cropped.tif
ERROR:root:Unhandled exception:
Traceback (most recent call last):
  File "C:\Users\rober\miniconda3\envs\clustering\lib\site-packages\napari\_qt\event_loop.py", line 79, in gui_qt
    yield app
  File "start_assistant.py", line 12, in <module>
    assistant_gui = napari_pyclesperanto_assistant.napari_plugin(viewer)
  File "c:\structure\code\napari_pyclesperanto_assistant\napari_pyclesperanto_assistant\_gui\_napari_plugin.py", line 6, in napari_plugin
    gui = AssistantGUI(viewer)
  File "c:\structure\code\napari_pyclesperanto_assistant\napari_pyclesperanto_assistant\_gui\_AssistantGui.py", line 26, in __init__
    self._init_gui()
[...]

PR #83, opened 2021-01-05
Clarify choices callable deprecation warning

tlambert03:
The deprecation warning provided when a widget receives a `choices` option that is a callable using the pre 0.2.0 API is confusing.  The "main" user of providing a function as widget choices is probably napari's [`napari.utils._magicgui.get_layers`](https://github.com/napari/napari/blob/master/napari/utils/_magicgui.py#L80) function, which is registered with `magicgui.register_type` [right here](https://github.com/napari/napari/blob/a0b38907487d8bd32d2a7300bcbbecfcd5b00b40/napari/utils/_magicgui.py#L39).  As such, anyone using napari <0.4.3 with magicgui >= 0.2.0 will get a warning that is not their fault (see #81):

```
As of magicgui 0.2.0, a `choices` callable may accept only a single positional
argument (an instance of `magicgui.widgets.CategoricalWidget`), and must return
an iterable (the choices to show). Function 'get_layers' accepts 2 arguments.
In the future, this will raise an exception.
```

This PR catches that outdated earlier (when actually registering the type), provides only a `DeprecationWarning` that most users won't see

```
DEVELOPER NOTICE: As of magicgui 0.2.0, when providing a callable to `choices`, the
callable may accept only a single positional argument (which will be an instance of
`magicgui.widgets._bases.CategoricalWidget`), and must return an iterable (the choices
to show).  Function 'napari.utils._magicgui.get_viewers' accepts 2 arguments.
In the future, this will raise an exception.
```

and then patches the function to match the new API.

### DIFF
--- a/magicgui/type_map.py
+++ b/magicgui/type_map.py
@@ -4,6 +4,7 @@ import inspect
 import pathlib
 import sys
 import types
+import warnings
 from collections import abc, defaultdict
 from enum import EnumMeta
 from typing import Any, DefaultDict, Dict, ForwardRef, List, Optional, Tuple, Type, cast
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
LGTM!

PR #86, opened 2021-01-05
add call count to function gui

tlambert03:
Add a call-count to FunctionGui.  @haesleinhuepf  could use this in `napari_pyclesperanto_assistant` [here](https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/851bb19b5190cc87477972f1fdb867f0f09ac066/napari_pyclesperanto_assistant/_gui/_LayerDialog.py#L25)


sofroniewn (approved):
Wow, nice! This is very cool!

PR #92, opened 2021-01-11
Extend combobox api with set_choice, get_choice, del_choice

tlambert03:
This PR will help to fix an issue in napari where the data for a single choice become stale, but `refresh_choices` is not necessarily desirable, as it will update all choices in a CategoricalWidget.
It adds a few methods to get/set/delete the data for a single choice.

I'm considering, in the near future, making the categorical widget derive from `MutableMapping` (so it would behave like a dict where the keys are the strings in the combobox, and the values are the data associated with each).  so I'm not terrible concerned about the API yet.  This solves the immediate issue at hand

### DIFF
--- a/magicgui/widgets/_bases.py
+++ b/magicgui/widgets/_bases.py
@@ -736,6 +736,30 @@ class CategoricalWidget(ValueWidget):
         self.choices = self._default_choices
 
     @property
+    def current_choice(self) -> str:
+        """Return the text of the currently selected choice."""
+        return self._widget._mgui_get_current_choice()
+
+    def __len__(self) -> int:
+        """Return the number of choices."""
+        return self._widget._mgui_get_count()
+
+    def get_choice(self, choice_name: str):
+        """Get data for the provided ``choice_name``."""
+        self._widget._mgui_get_choice(choice_name)
+
+    def set_choice(self, choice_name: str, data: Any = None):
+        """Set data for the provided ``choice_name``."""
+        data = data if data is not None else choice_name
+        self._widget._mgui_set_choice(choice_name, data)
+        if choice_name == self.current_choice:
+            self.changed(value=self.value)
+
+    def del_choice(self, choice_name: str, data: Any = None):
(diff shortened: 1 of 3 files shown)


magicgui/widgets/_bases.py
```diff
@@ -735,6 +735,30 @@ def reset_choices(self, event=None):
...
+    @property
+    def current_choice(self) -> str:
+        """Return the text of the currently selected choice."""
+        return self._widget._mgui_get_current_choice()
+
+    def __len__(self) -> int:
+        """Return the number of choices."""
+        return self._widget._mgui_get_count()
+
+    def get_choice(self, choice_name: str):
```
sofroniewn, line 747: Could this be `get_choice_data` or is that not quite right?

magicgui/widgets/_bases.py
```diff
@@ -735,6 +735,30 @@ def reset_choices(self, event=None):
         """
         self.choices = self._default_choices
 
+    @property
+    def current_choice(self) -> str:
```
sofroniewn, line 739: Could this be `current_choice_name`, or do you get `data` back too?

sofroniewn (approved):
Looks good, I left a minor comment on naming. I don't think it's important to worry about now, but it is a little confusing with `name` and `data`.

tlambert03:
I totally agree. There's too many names for the same thing floating around.  You're correct in both cases.  This is one reason why I'm kinda thinking about using `MutableMapping` (with the addition that there is a "`current_key`", and `value` is a shortcut for `widget.get(widget.current_key)`):
```python
widget.update({'a': 1, 'b': 2, 'c': 3})   # dropdown shows a, b, & c
widget['d'] = 4   # a fourth item is added
widget.current_key = 'b'  # changes the selection
widget.value  # == 2

list(widget.items())  # [('a', 1), ('b', 2) ...
```

tlambert03:
lets punt this discussion for another release

sofroniewn:
> lets punt this discussion for another release

sounds good

PR #95, opened 2021-01-13
better bound values

tlambert03:
This PR provides much better support for "binding" a value to a given widget (effectively circumventing `widget.value` when `widget._bound_value` is not `None`).  The bound value can also be a callable (that accepts a single param), in which case accessing `widget.value` will really return `widget._bound_value(widget)`.

All told, this makes it trivial for other libraries (like napari) to bind a value or callback to a parameter with a particular annotation (e.g. `def function(viewer: napari.Viewer)`) ... letting that parameter be hidden from the widget, but still present when calling the function.

Bound parameters are hidden by default, but can be shown explicitly with `{'visible': True}`.

For examples of how this can be used, see [tests here](https://github.com/napari/magicgui/pull/95/files#diff-af3529af46c0c6d7c1c94af472ea2a6b21b19cc715c2c7207e53f1c8f7fc9bdaR219-R266).

*edit: this also provides a much more elegant mechanism to decorate class methods #56  ... and undoes most of that PR*

### DIFF
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -217,17 +217,35 @@ def test_unhashable_choice_data():
 
 
 def test_bound_values():
+    """Test that we can bind a "permanent" value override to a parameter."""
+
     @magicgui(x={"bind": 10})
     def f(x: int = 5):
         return x
 
+    # bound values hide the widget by default
     assert not f.x.visible
     assert f() == 10
     f.x.unbind()
     assert f() == 5
 
 
+def test_bound_values_visible():
+    """Test that we force a bound widget to be visible."""
+
+    @magicgui(x={"bind": 10, "visible": True})
+    def f(x: int = 5):
+        return x
+
+    assert f.x.visible
+    assert f() == 10
+    f.x.unbind()
+    assert f() == 5
+
+
 def test_bound_callables():
+    """Test that we can use a callable as a bound value."""
+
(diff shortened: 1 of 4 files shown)


tests/test_widgets.py
```diff
@@ -214,3 +214,53 @@ def test_unhashable_choice_data():
...
+
+def test_bound_values():
+    @magicgui(x={"bind": 10})
+    def f(x: int = 5):
+        return x
+
+    assert not f.x.visible
+    assert f() == 10
+    f.x.unbind()
+    assert f() == 5
```
sofroniewn, line 227: Is it worth adding one more test with the `visible=True` case. Did I get the syntax right?

```python
def test_bound_values_visible():
    @magicgui(x={"bind": 10, "visible": True})
    def f(x: int = 5):
        return x

    assert f.x.visible
    assert f() == 10
    f.x.unbind()
    assert f() == 5
```

sofroniewn (approved):
LGTM! Makes a lot of sense and will be nice for the napari usecase!

tests/test_widgets.py
tlambert03: yes indeed!  will do

tlambert03: added

sofroniewn: thanks!

PR #96, opened 2021-01-13
Fix reset_choices recursion

tlambert03:
This fixes a bug in reset_choices recursion, when a FunctionWidget is nested inside of a Container


sofroniewn (approved):
Good catch! LGTM!

PR #100, opened 2021-01-14
Add manual and docstring-parsed tooltips

### LINKED ISSUE #94: Tooltips
Widgets should be able to have tooltips.  My first thought is to parse the docstring and take part (or all?) of the parameter description for each parameter, along with, of course, accepting a `tooltip: str= ..` parameter directly on the widget.

tlambert03:
This closes #94 by adding an option to provide a `tooltip` parameter to any widget.  When using with a `magicgui` decorator, tooltips will be parsed (by default) from the first sentence of the corresponding parameter description in the docstring (ReST, google and numpy docstrings supported).  Auto-tooltips can be turned off using `magicgui(tooltips=False)`, or overridden for each parameter specifically.

```python
# during widget creation:
widget = LineEdit(tooltip="Enter a value")

# with magicgui
@magicgui(x={"tooltip": "override tooltip"}, z={"tooltip": None})
def func(x: int, y: str = "hi", z=None):
    """Do a little thing.

    Parameters
    ----------
    x : int
        An integer for you to use
    y : str, optional
        A greeting, by default 'hi'. Only the first sentence
        will be used. Even if the docstring is very long, etc...
    z : Any, optional
        No tooltip for me please.
    """
    pass

assert func.x.tooltip == "override tooltip"
assert func.y.tooltip == "A greeting"  # the "by default" part is also stripped
assert not func.z.tooltip
```


magicgui/widgets/_bases.py
```diff
@@ -293,28 +294,6 @@ def native(self):
         """Return native backend widget."""
         return self._widget._mgui_get_native_widget()
 
-    def show(self, run=False):
```
tlambert03, line 296: most of this is just moved in this PR... only new things is the `tooltip` property

tlambert03:
@haesleinhuepf, you had expressed an interest in this.  have a look at the API above and let me know if you have any thoughts

jni (approved):
❤️

Issue #25, opened 2020-07-01
Progress bar widget for long-running computations


GenevieveBuckley:
Have you thought much about whether it might be possible to add a progress bar that updates with long-running computations? 

Most of the current examples (snell's law, etc.) are fast by design but I imagine many real-word cases might be much slower, and it might be good for users to have some kind of indication that things are running. This might get complicated because you'd need analysis functions to have a way to communicate back to the magicgui dockable widget. I wanted to get your thoughts given the work you did on [two way communication with threads in napari](https://github.com/napari/napari/pull/1258).

Relevant links:
* https://doc.qt.io/qt-5/qprogressbar.html

jni:
Well, one issue is that many computations don't really provide this, and we want to support arbitrary computations. Dask *does* provide one, so it's probably doable to hook into it.

But if you're saying, provide a widget that you can *write* a computation specifically to hook into with progress... That should be doable I think?

Anyway, certainly would be interesting... :grimacing:

tlambert03:
I agree, that would be cool feature.  But as @jni said, it's not clear how to achieve this in a generic way.  We could _create_ an API that a user could use (with generators and yields) that could do something like this, but i can't see my way to a solution without some degree of buy-in from the end-user (which might be fine).  For a hot minute, there was the [ProgressWorker](https://github.com/napari/napari/blob/3c80a7c2a99d4e1308f33d1c582225bbd2406f79/napari/_qt/threading.py#L346-L413) example in napari/napari#1210, which is the kind of thing I would imagine is needed.  but maybe we can think of a clean API to mimic the example progress bar in https://github.com/napari/napari/blob/master/examples/multithreading_two_way.py

GenevieveBuckley:
> But if you're saying, provide a widget that you can _write_ a computation specifically to hook into with progress... That should be doable I think?

Yes, I mean that the user would be the one to make sure their computation hooks into an API for indicating progress.

tlambert03:
With #104 and #105 in I'm going to close this.  While we haven't implemented any sort of threading support, the tqdm progress bar works reasonably well in a single threaded system, since it calls `app.processEvents()` on each iteration.  Threading can wait

PR #105, opened 2021-01-17
Progress bar tqdm wrapper, and manual control

tlambert03:
this PR adds a basic `tqdm_mgui` iterator that updates a progress bar (from #104), with the internal state of the tqdm iterator.  allowing iterator-based progress bars ( a step towards #25)
```python 
for i in tqdm_mgui(range(100)):
    time.sleep(0.1)
```

https://user-images.githubusercontent.com/1609449/104830197-688fa280-584a-11eb-94b6-f90adeb71406.mov


tlambert03:
this is getting kinda cool!
If you use `tqdm_mgui` (or the convenience `tmgrange` shortcut) inside of a `@magicgui`-decorated function, it will add the progressbar to the gui.  It also supports nested progressbars.  For instance:

```python
import random
from time import sleep
from magicgui import magicgui
from magicgui.tqdm import tmgrange, tqdm_mgui

@magicgui(call_button=True, layout="vertical")
def long_function(
    steps=20, repeats=4, choices="ABCDEFGHIJKLMNOP12345679", char="", delay=0.1
):
    # using the `tmgrange` shortcut instead of range()
    for r in tmgrange(repeats, label="repeats"):
        letters = [random.choice(choices) for _ in range(steps)]
        # `tqdm_gui`, like `tqdm`, accepts any iterable
        # this progress bar is nested and will be run & reset multiple times
        for letter in tqdm_mgui(letters, label="steps"):
            long_function.char.value = letter
            sleep(delay)

long_function.show(run=True)
```


https://user-images.githubusercontent.com/1609449/104848531-5e55be80-58b3-11eb-8dca-69b1a4629dd6.mov

tlambert03:
cc @GenevieveBuckley

tlambert03:
progress bar can now also be controlled by annotating a parameter as `magicgui.widgets.ProgressBar`:
```python
@magicgui(call_button="tick", pbar={"min": 0, "step": 2, "max": 20, "value": 0})
def manual(pbar: ProgressBar, increment: bool = True):
    """Example of manual progress bar control."""
    if increment:
		# can also use pbar.value, pbar.min, pbar.max, etc...
        pbar.increment()
    else:
        pbar.decrement()
```

sofroniewn:
Wow, this is looking really cool!! I can give a proper review later, but overall the functionality looks great. I'll need to do a little investigation to understand when I'd use `tqdm_gui`, `tmgrange`, and `magicgui.widgets.ProgressBar`. I've not used `tqdm` itself much, so maybe that's why it's not immediately obvious to me, but I'm excited to give this a go!

magicgui/tqdm.py
```diff
@@ -0,0 +1,151 @@
+"""A wrapper around the tqdm.tqdm iterator that adds a ProgressBar to a magicgui."""
```
sofroniewn, line 1: Just wondering if there is another place this could live that is not at the top level, i.e. a grouping of a larger thing that it might be an example of? I guess though there isn't anything else quite like in magicgui right now as it's not a widget per se (that is `ProgressBar` itself, this is the iterator) but something that if found inside magic functions with generate a widget?

sofroniewn, line 1: Just wondering if there is another place this could live that is not at the top level, i.e. a grouping of a larger thing that it might be an example of? I guess though there isn't anything else quite like in magicgui right now as it's not a widget per se (that is `ProgressBar` itself, this is the iterator) but something that if found inside magic functions with generate a widget?

sofroniewn, line 1: Just wondering if there is another place this could live that is not at the top level, i.e. a grouping of a larger thing that it might be an example of? I guess though there isn't anything else quite like in magicgui right now as it's not a widget per se (that is `ProgressBar` itself, this is the iterator) but something that if found inside magic functions will generate a widget?

examples/progress.py
```diff
@@ -0,0 +1,61 @@
...
+# `tmgrange(*args, **kwargs)` is a shortcut for tqdm_mgui(range(*args), **kwargs)
+# (..., leave=False) will make the progressbar hide when done
+for i in tmgrange(10, leave=False):
+    sleep(0.05)
+
+
+# can also be used inside of a magicgui decorator
+# in which case the progress bar will be added to the magicgui container
+@magicgui(call_button=True)
+def long_running(steps=10, delay=0.1):
```
sofroniewn, line 18: If functions with `tmgrange` or `tqdm_mgui` inside then don't get magic gui decorated will they just behave as normal with the default behavior of of `tqdm`?

magicgui/tqdm.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+_tqdm_kwargs = {
+    p.name
+    for p in inspect.signature(tqdm.__init__).parameters.values()
+    if p.kind is not inspect.Parameter.VAR_KEYWORD and p.name != "self"
+}
+
+
+class tqdm_mgui(tqdm):
```
sofroniewn, line 43: Curious why not just have this class be called `tqdm`, to use it you're importing `from magicgui import tqdm` already, especially if this is a drop in replacement for `tqdm` / behavior is identical if not inside a magicgui decorated function

magicgui/tqdm.py
```diff
@@ -0,0 +1,151 @@
...
+
+            if not self.leave:
+                self._app.process_events()
+                self.progressbar.hide()
+
+        if self._mgui:
+            self._mgui._tqdm_depth -= 1
+
+
+def tmgrange(*args, **kwargs) -> tqdm_mgui:
```
sofroniewn, line 149: I guess this is analogous to `trange`, might be nice to link to that https://tqdm.github.io/docs/tqdm/#trange somehow. I'm also wondering if the above import were to change to `from magicgui import tqdm` would this be `from magicgui import trange`, maybe not, that's a little awkward?

sofroniewn (commented):
Ok, this is looking good i left some comments/ questions, most are for my clarification/ naming ideas that could be ignored - probably the only thing i see as critical is enable/ disable idea for fixing the segfault after clicking run twice.

magicgui/tqdm.py
sofroniewn: duplicate of abvoe

sofroniewn: ha! triplicate of above, my github was clearly acting up!!!

examples/progress.py
```diff
@@ -0,0 +1,61 @@
+import random
```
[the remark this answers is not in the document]
tlambert03: can do! thanks

examples/progress.py
tlambert03: at the moment no, they will also pop up a little GUI progress bar (it just won't be in any parent widget)... but that is absolutely possible and definitely a goal (i.e. it should be smart enough to detect the context in which it is called)

examples/progress.py
```diff
@@ -0,0 +1,61 @@
...
+# can also be used inside of a magicgui decorator
+# in which case the progress bar will be added to the magicgui container
+@magicgui(call_button=True)
+def long_running(steps=10, delay=0.1):
+    """Long running computation with range iterator."""
+    for i in tmgrange(steps):
+        sleep(delay)
+
+
+long_running.show(run=True)
```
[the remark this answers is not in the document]
tlambert03: yeah I noticed that too.  That's a larger issue having to do with long running functions that doesn't have to do with this PR ... will make a new issue

magicgui/tqdm.py
tlambert03: > I guess though there isn't anything else quite like in magicgui right now as it's not a widget per se 

exactly.  this is just the iterator magic.  the main reason I put it at the top level is that I don't want to actually depend on `tqdm` directly (it's available as `magicgui[tqdm]`), and I want people to use `from magicgui.tqdm import tqdm...`.  As you point out, you can still use the manual progressbar API without this (magicgui.widgets.ProgresBar)

magicgui/tqdm.py
tlambert03: because it's not yet a drop-in replacement :/

magicgui/tqdm.py
tlambert03: see above about not depending directly on tqdm.  I wanted the import error to be raised on accessing `magicgui.tqdm`

tlambert03 (commented):
thanks a lot!

magicgui/tqdm.py
tlambert03: ok now it is!  I liked this idea... so now it's just `from magicgui.tqdm import tqdm, trange` ... and if it's not inside of a `@magicgui` it will work as usual (like in IPython)

examples/progress.py
tlambert03: fixed... now it will just be the default behavior

examples/progress.py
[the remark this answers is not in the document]
tlambert03: done, thanks

examples/progress.py
sofroniewn: nice! i think that's pretty cool!!! enables maximum reusability

examples/progress.py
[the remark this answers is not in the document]
sofroniewn: sounds good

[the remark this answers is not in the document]
sofroniewn: that issue might need to get fixed first, but thinking about it more I think i'd like the behavior to toggle `run/ pause` which would start and stop progress. could come in later PRs, but noting it now

magicgui/tqdm.py
sofroniewn: love it so much!!

examples/progress.py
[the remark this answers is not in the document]
sofroniewn: thanks, i find it much easier to parse now

sofroniewn (approved):
Ok, I love it - looks great to me. I made an additional comment on the `run/ pause` toggle, but sounds like that might have to come in a new PR. I think that's fine, but depending on how much work it is might try and get it in before the next release too. I find it very tempting to click on the run button again and imagine others will too :-)

jni (approved):
@tlambert03 pretty happy with this. I'm not 100% about "overwriting" the tqdm name, but I see the argument in favour, so let's give it a go.

Overall, I was kind of envisioning a different, more magical API using generators. Half baked idea:

```python
@magicgui(
    progress=[argname],  # where to get the total length
)
def generator(argname : List):
    for elem in list:
        process(elem)
        yield
    return
```

There's every chance this current API is better though; For one, my suggestion can't do nested iterations, or resetting, or any of that. But, maybe it could be extended (multiple args, yielded values). At any rate, these are all things that can be explored in the future.

tlambert03:
> Overall, I was kind of envisioning a different, more magical API using generators. Half baked idea:

(as discussed on zulip), I like this idea too, and agree that it isn't mutually exclusive.  I like that the current pattern lets someone simply wrap the iterator itself (i.e. it doesn't require you to specify the parameters of the iterable in two places), but I also like thinking about what we could do with generators.  so I'll merge this, and keep thinking about that as a complementary pattern

Issue #107, opened 2021-01-18
Best way to add general docs and help menu?


HagaiHargil:
Hi :)

My main use case for magicgui is to replace command line applications with friendlier, GUI-based tools. This is very helpful for my non-coder colleagues, and magicgui has been invaluable on that front. One thing that I "miss" from CLI tools is documentation for the available parameters and more general "help documentation" which explains the use case that this current app is here to solve. In CLI tools these bits of information can usually be invoked with a `--help` flag which is obviously missing here.

I was wondering whether you think it would be possible to add text box for each parameter of the GUI which appears following a hover "event". This text box may pull its information from the docstring of the main annotated function (moar magik!). If you think it's doable I might try to dip my toes into this magical swamp.

In addition, a more traditional "Help" context menu or button can be optionally activated by adding a `"help_button: True"` entry to the main magicgui dictionary. This button can again get its information from the main body of the function's docstring.

How does it sound to you? I believe some docstring magic is already happening in order to decipher each type's designated widget.

tlambert03:
> How does it sound to you?

sounds great!  it's definitely inline with the goal of making it easy for developers to provide _usable_ GUIs to people (and if it's not clear what all the widgets do, then it's not very useable! :joy:)

> I believe some docstring magic is already happening in order to decipher each type's designated widget.

we actually don't really use the docstrings much for type or widget inference (that all comes from either the type annotation, the default value, or the parameters to `magicgui()`)... however, you might be interested in the tooltips PR that recently merged! https://github.com/napari/magicgui/pull/100  (not yet released)

That should hopefully satisfy _part_ of the need here... but as you'll notice in that PR, the tooltips are restricted to the first sentence of the parameter docstring.  So a full-blown help would also be useful (maybe if `help_buton` is true, a little ℹ️ could appear somewhere in the widget that pops up the full docstring?)

tlambert03:
I haven't added tooltips to the online docs yet... so until then, the best examples for tuning its behavior are probably [these tests](https://github.com/napari/magicgui/blob/45e7a15ac85f5e9a12249575c8efff80e0172d34/tests/test_magicgui.py#L483-L589)

HagaiHargil:
Awesome, I missed that issue and the PR completely! I knew that "hover text box" wasn't the right phrase but I couldn't find a better one, thanks for these references.

I wonder why do you parse only the first sentence (up to the first comma or period)? Is there some technical limitation or do you think that it's simply better this way?

[...]

PR #108, opened 2021-01-18
Parse the entire docstring for the tooltip

HagaiHargil:
Following #107, I removed the lines that cut off the docstring parsing and fixed the tests. Are there any more places that require an update?


tlambert03 (approved):
thanks!  looks good to me

PR #110, opened 2021-01-20
Add `MainWindow` variant on `Container`, and `MainFunctionGui`

tlambert03:
@HagaiHargil, here's a start. This adds a new `MainWindow` widget type, that has one (currently) new method: `self.create_menu_item("Menu Name", "Menu Item Name", callback=some_callback_when_selected, shortcut='keyboard shortcut')`

if you use `magicgui(main_window=True)` you get one with the docstring in the help menu.  give it a try (see `examples/widget_demo.py`) and let me know if this general idea might work for you


HagaiHargil:
This is awesome! Thank you so much. One quirk - for some reason I can edit the text of the "Documentation" pop up text box. But other than that it's perfect.

tlambert03:
ok great!  Just wanted to make sure we're on the right track... Next I can make the documentation window look better, and un-editable.

I was thinking of perhaps parsing the docstrings and formatting as html, do you like that?  Or shall we just leave as plaintext

HagaiHargil:
Well now that you mention it... :D
If an HTML document can be made out of the documentation then I can definitely see myself adding illustrations and images showing how this current GUI\tool should be used.

It will definitely be nice to have, but not a top priority.

HagaiHargil:
Very nice! I believe that the main docstring is in italics for some reason. Do you see it as well?

Also, would you like me to add examples, docs, use cases, whatever?

tlambert03:
> Very nice! I believe that the main docstring is in italics for some reason. Do you see it as well?

fixed

> Also, would you like me to add examples, docs, use cases, whatever?

that would be super helpful!  Perhaps another example (I'm using `widget_demo.py`, but a more topic-specific demo would be great).  And if you're up for it, a new doc page at either `docs/usage/main_window.md` or `docs/examples/main_window.md` would be great.  You can use regular markdown (probably no need to mimic the jupytext stuff used in some of the other docs markdowns since we can't actually render the menu in the docs), so just regular fenced code examples would be great.

Maybe make a new PR for us to discuss it, though I'd like to add some more tests here before merging this one, is that ok?

HagaiHargil:
Sure thing.

PR #117, opened 2021-01-21
Magic factory

tlambert03:
As discussed on zulip, this adds a `@magic_factory` decorator that, instead of returning a `FunctionGui`, returns a `MagicFactory` instance (basically just a partial subclass that we can type check) which will return the actual `FunctionGui` instance when called.

still need to add tests and take another pass through it, but this can be evaluated if you want @jni

**update (1/21)**

Here is where the API has settled and what it can do:

```python
from magicgui import magicgui, magic_factory
from magicgui.widgets import FunctionGui, Slider, SpinBox
from magicgui._magicgui import MagicFactory

# the magicgui decorator creates a single widget instance
@magicgui
def func():
    pass

assert isinstance(func, FunctionGui)

# the magic_factory decorator creates a factory function
# that must be called to produce the widget

@magic_factory
def factory(x: int):
    pass

assert isinstance(factory, MagicFactory)
widget = factory()
assert isinstance(widget, FunctionGui)
assert isinstance(widget.x, SpinBox)

# While the `MagicFactory` will remember the kwargs provided
# to the `magic_factory` decorator, you can also override them
# individually when creating the widget

widget2 = factory(x={'widget_type': 'Slider'})
assert isinstance(widget2.x, Slider)
```

self-referencing _also_ works with the `@magic_factory` decorator (meaning that inside the body of the function, the name of the function will be pointing to the ultimate widget instance)... with the main caveat being that this only works when decorating a module-level function:

```python
# this works

@magic_factory
def factory(x: int):
    assert isinstance(factory, FunctionGui)
    print(factory)

widget = factory()
widget()  # prints "<FunctionGui factory(x: int = 0)>"
```

this, however, does _not_ work:

```python
def outer():
    @magic_factory
    def inner(x: int):
        assert isinstance(inner, FunctionGui)
        print(factory)
    return inner

factory = outer()
assert isinstance(factory, MagicFactory)
widget = factory()
widget()  # AssertionError
```

personally, I'm just fine with this caveat.


magicgui/widgets/_function_gui.py
```diff
@@ -105,7 +105,7 @@ def __init__(
         if tooltips:
             _inject_tooltips_from_docstrings(function.__doc__, param_options)
 
-        self._function = function
+        self.__wrapped__ = function
```
tlambert03, line 108: changing this to `__wrapped__` for potential future compatibility with https://docs.python.org/3/library/functools.html

jni:
Ha, the failing tests in 3.7:

```
  E     File "/home/runner/work/magicgui/magicgui/magicgui/_magicgui.py", line 27
  E       def __call__(self, /, *args, **keywords) -> FunctionGui:
  E                          ^
  E   SyntaxError: invalid syntax
```

You're just too damn clever, @tlambert03! 😂 At any rate that seems like a super unnecessary use of `/`?

jni:
@tlambert03 as I started pondering about this in my head, I had the dreadful feeling that it wouldn't work, and indeed that feeling has been realised:

```pytb
    mode = start_affinder._call_button.text  # can be "Start" or "Finish"
AttributeError: 'MagicFactory' object has no attribute '_call_button'
```

All of the "reflexive" tricks we do no longer work... Not sure how to deal with it though. =\

tlambert03:
> You're just too damn clever, @tlambert03! 😂 At any rate that seems like a super unnecessary use of `/`?

Hah, good point, I actually just copied that straight from `functools.partial.__call__` but forgot about that!

tlambert03:
> All of the "reflexive" tricks we do no longer work... Not sure how to deal with it though. =\


Ohhhh nooo :( forgot about that stuff.  My first thought is that maybe we try to use a modified locals() in magic_factory, that forwards just the name of the function?  (But haven't tried to actually do that)

tlambert03:
ok, @jni, I think I've got a solution I'm happy with, read the update in the main post for details... the main point is that self-reference will now work fine for factories decorated at the module level, but not for factories decorated in local scopes.  (which is fine with me... I'm not going to bother at the moment detecting that probably-rare scenario to raise a warning).

If you want to see the guts of how I did that see the context manager used [here](https://github.com/tlambert03/magicgui/blob/a0c08d98413d6c02c7aeba3a8901bfb475cdd08e/magicgui/widgets/_function_gui.py#L228-L229) and defined [here](https://github.com/tlambert03/magicgui/blob/a0c08d98413d6c02c7aeba3a8901bfb475cdd08e/magicgui/widgets/_function_gui.py#L330-L332)

tlambert03:
@brisvag, you may also want to look at the API in the updated first post. Let me know if you have suggestions or questions.

brisvag:
That looks great! Overriding kwargs is perfect :+1:

What I would like to be able to do, and tried to (poorly) describe on zulip was something a bit more complex: accepting arbitrary amounts and combinations of kwargs at *runtime*. After thinking about it a bit more, I think it's a very different thing from what's happening here, and might be entirely different from the goal of magicgui. Or it might be already possible, and I'm thinking about it the wrong way.
Here's some mockup of what I envisioned:

```python
@magicgui
def factory(**kwargs: Union[int, float]):
    for k, v in kwargs.items():
        if isinstance(v, int):
            # do something
        if isinstance(v, float):
            # do something else

factory(x=1, y=2.0)
```
And this would generate a widget equivalent to:
```python
@magicgui
def factory(x: int, y: float):
    # stuff
```

tlambert03:
(chatted with @brisvag on zulip and have a better idea of how we might help there... probably not the factory)

tlambert03:
> I'm not going to bother at the moment detecting that probably-rare scenario to raise a warning

oh I bothered... got a nice warning now :)

magicgui/_magicgui.py
```diff
@@ -144,36 +187,176 @@ def magicgui(
...
+    @property
+    def __name__(self) -> str:
+        """Pass function name."""
+        return getattr(self.keywords.get("function"), "__name__", "FunctionGui")
+
+
+# Overloads for magic_factory decorator.  See implementation below
+# TODO: figure out how to get these in a stub file.
+# My first attempts to put them in ``_magicgui.pyi`` broke my type hints in VSCode
+# fmt: off
```
sofroniewn, line 261: Yeah, that might be nice. I just had to read up on [overloading again](https://www.python.org/dev/peps/pep-3124/#rationale-and-goals) and that seems like a good goal, but not critical for PR

sofroniewn (approved):
Ok this looks good to me, I havn't used it yet - will leave that to @jni, but it's clearly laid out and I think good to merge as is

magicgui/_magicgui.py
tlambert03: yeah, I know how ugly that all looks... sorry.  but i'm not willing to give up what it provides in the IDE:

```
@magic_factory
def func() -> int:
    return 1

widget = func()  # mypy knows widget is a FunctionGui instance
main = func(main_window=True) # mypy knows main is a MainFunctionGui instance
x = widget()  # mypy knows x is an integer
```
I looove that.  but using .pyi stubs isn't working for me yet

magicgui/_magicgui.py
```diff
@@ -1,18 +1,70 @@
...
+
+_T = TypeVar("_T")
+_R = TypeVar("_R")
+
+
+def _magicgui(function=None, factory=False, main_window=False, **kwargs):
+    """Actual private magicui decorator.
+
+    if factory is `True` will return a MagicFactory instance, that can be called
+    to return a `FunctionGui` instance.  See docstring of ``magicgui`` for parameters
```
jni, line 33:
```suggestion
    to return a `FunctionGui` instance.  See docstring of ``magicgui`` for parameters.
    Otherwise, this will return a FunctionGui instance directly.
```

magicgui/_magicgui.py
```diff
@@ -1,18 +1,70 @@
...
+    """
+    if "result" in kwargs["param_options"]:
+        warn(
+            "\n\nThe 'result' option is deprecated and will be removed in the future."
+            "Please use `result_widget=True` instead.\n",
+            FutureWarning,
+        )
+
+        kwargs["param_options"].pop("result")
+        kwargs["result_widget"] = True
```
jni, line 43: what did `result` do before? I'm confused about why this isn't `kwargs["result_widget"] = kwargs["param_options"].pop("result")`

jni: Oh, was "result" a special parameter name in magicgui? That's pretty terrible. LOL I might consider removing it altogether immediately. (But probably in a separate PR.) Generally, special names like that should probably have underscores or other prefixes so they don't conflict with actual parameters.

tlambert03: yeah that seems very reasonable!  I need to go back and look now, but it's possible it used to accept a string?

tlambert03: yeah, it was terrible! I agree!  I can remove in another PR

magicgui/_magicgui.py
```diff
@@ -1,18 +1,70 @@
...
+
+    def inner_func(func: Callable) -> Union[FunctionGui, MagicFactory]:
+        if not callable(func):
+            raise TypeError("the first argument must be callable")
+
+        magic_class = MainFunctionGui if main_window else FunctionGui
+
+        if factory:
+            return MagicFactory(func, magic_class=magic_class, **kwargs)
+        return magic_class(func, **kwargs)
```
jni, line 53: This logic I find confusing... Why isn't this (forgive the names):

```python
a_factory = MagicFactory(func, magic_class=magic_class, **kwargs)
if factory:
    return a_factory
else:
    return a_factory()
```

magicgui/_magicgui.py
```diff
@@ -1,18 +1,70 @@
...
+
+    if function is None:
+        return inner_func
+    else:
+        return inner_func(function)
```
jni, line 58: In what situations does `function=None` make sense? Is this that weird trick that could be avoided with toolz.curry? =)

magicgui/_magicgui.py
tlambert03: I guess I felt like there was no need to go through the whole `MagicFactory.__new__` instantiation and stuff if we knew we were immediately going to just the instance.

magicgui/_magicgui.py
tlambert03: yeah, we've had this discussion before! 😂  ... but I don't want to depend on toolz

tlambert03: ... and `function=None` makes sense when you provide parameters to the decorator:
```python
@magicgui(call_button=True)  # magicgui is being called with function=None
def function():
    ...
```

magicgui/_magicgui.py
```diff
@@ -23,13 +75,10 @@ def magicgui(  # noqa
     main_window: Literal[False] = False,
     app: AppRef = None,
     **param_options: dict,
-) -> FunctionGui:
-    ...
-
-
-@overload
+) -> FunctionGui[_R]: ...
```
jni, line 78: Pretty sweet use of typevars! =)

jni: I mean, I know this is exactly the point of them :joy:, I'd just never seen them in the wild. =)

tlambert03: it _is_ lovely to work with! (now if I can only figure out how to move it all to `_magicgui.pyi`

magicgui/_magicgui.py
```diff
@@ -144,36 +187,176 @@ def magicgui(
...
+                "MagicFactory missing required positional argument 'function'"
+            )
+        if isinstance(function, FunctionType):
+            if "<locals>" in function.__qualname__:
+                freevars = function.__code__.co_freevars
+                if function.__name__ in freevars:
+                    warn(
+                        "Self-reference detected in MagicFactory function created "
+                        "in a local scope. FunctionGui references will not work."
+                    )
```
jni, line 224: This is just so insane. :smile:

jni: Indeed, I would suggest adding a comment above the `freevars = ` line to explain wth you are doing here. I would have exactly *zero* idea if I hadn't been the cause of this thing. :joy:

magicgui/_magicgui.py
```diff
@@ -144,36 +187,176 @@ def magicgui(
...
+
+    def __repr__(self) -> str:
+        """Return string repr."""
+        params = inspect.signature(magicgui).parameters
+        args = [
+            f"{k}={v!r}"
+            for (k, v) in self.keywords.items()
+            if v not in (params[k].default, {})
+        ]
+        return f"MagicFactory({', '.join(args)})"
```
jni, line 237: Jesus. :joy:

magicgui/_magicgui.py
tlambert03: image

??

magicgui/_magicgui.py
```diff
@@ -144,36 +187,176 @@ def magicgui(
...
+
+# Overloads for magic_factory decorator.  See implementation below
+# TODO: figure out how to get these in a stub file.
+# My first attempts to put them in ``_magicgui.pyi`` broke my type hints in VSCode
+# fmt: off
+@overload  # noqa: E302
+def magic_factory(  # noqa
+    function: Callable[..., _R],
+    *,
+    layout: str = "horizontal",
```
jni, line 266: Have we not settled on vertical default yet?

magicgui/_magicgui.py
```diff
@@ -144,36 +187,176 @@ def magicgui(
...
+def magic_factory(  # noqa
+    function: Literal[None] = None,
+    *,
+    layout: str = "horizontal",
+    labels: bool = True,
+    tooltips: bool = True,
+    call_button: Union[bool, str] = False,
+    auto_call: bool = False,
+    result_widget: bool = False,
+    main_window: Literal[True],
```
jni, line 314: Q for my own education: does Literal work if it's not actually a literal? ie will this code pass type checking?

```python
main_win = True
f = magic_factory(func, main_window=main_win)()
f.<something_main_function_gui_provides>()
```
?

jni: Or even worse, something where `main_win` is determined from external input...

magicgui/_magicgui.py
```diff
@@ -144,36 +187,176 @@ def magicgui(
...
+    result_widget: bool = False,
+    main_window: bool = False,
+    app: AppRef = None,
+    **param_options: dict,
+):
+    """Return a :class:`MagicFactory` for ``function``."""
+    return _magicgui(factory=True, **locals())
+
+
+_factory_doc = magicgui.__doc__.split("Returns")[0] + (  # type: ignore
```
jni, line 338: You are a madman, @tlambert03! :joy:

magicgui/widgets/_function_gui.py
```diff
@@ -109,13 +117,24 @@ def __init__(
...
+        # it's conceivable that function is not actually an instance of FunctionType
+        # we can still support any generic callable, but we need to be careful not to
+        # access attributes (like `__name__` that only function objects have).
+        # Mypy doesn't seem catch this at this point:
+        # https://github.com/python/mypy/issues/9934
+        self._callable_name = (
+            getattr(function, "__name__", None)
+            or f"{function.__module__}.{function.__class__}"
+        )
```
jni, line 129: Nice!

magicgui/_magicgui.py
tlambert03: not yet, but I could be convinced.

magicgui/widgets/_function_gui.py
```diff
@@ -169,7 +188,9 @@ def __getattr__(self, value):
     #     """Delete a widget by integer or slice index."""
     #     raise AttributeError("can't delete items from a FunctionGui")
 
-    def __call__(self, *args: Any, **kwargs: Any):
+    _UNSET = object()  # sentinel used when temporarily overriding function.__globals__
```
jni, line 191: Can you elaborate on this comment?

magicgui/_magicgui.py
tlambert03: yep, that works

magicgui/_magicgui.py
tlambert03: gotta be DRY!

magicgui/widgets/_function_gui.py
```diff
@@ -344,3 +366,35 @@ def _docstring_to_html(docs: str) -> str:
...
+
+
+@contextmanager
+def _function_name_pointing_to_widget(function_gui: FunctionGui):
+    """Context in which the name of the function points to the function_gui instance.
+
+    When calling the function provided to FunctionGui, we make sure that the name
+    of the function points to the FunctionGui object itself (this object).
+    In standard ``@magicgui`` usage, this will have been the case anyway.
+    Doing this here allows @magic_factories to *also* be self-referential.
```
jni, line 381: Maybe link to the various comments/conversations about this? Or add a minimal example of self-referencing here?

magicgui/widgets/_function_gui.py
```diff
@@ -344,3 +366,35 @@ def _docstring_to_html(docs: str) -> str:
...
+    # other functions.
+    original_value = function.__globals__.get(func_name, _UNSET)
+    function.__globals__[func_name] = function_gui
+    try:
+        yield
+    finally:
+        if original_value is _UNSET:
+            del function.__globals__[func_name]
+        else:
+            function.__globals__[func_name] = original_value
```
jni, line 400: Remarkable. :grinning:

jni (approved):
@tlambert03 all of my comments are for clarification, I'll let you use your judgement regarding which/how to address. Amazing PR. :grinning:

magicgui/widgets/_function_gui.py
tlambert03: oops, sorry, I forgot about that one when i moved it down to the bottom of the file.  Removed

magicgui/widgets/_function_gui.py
tlambert03: expanded comment with example

magicgui/_magicgui.py
tlambert03: #122

magicgui/_magicgui.py
tlambert03: #121

magicgui/_magicgui.py
jni: Right! Maybe add a comment like, "magic factory needs to be a lot more magical than necessary if we are instantiating straight away, so we shortcut that here"?

magicgui/_magicgui.py
jni: Fine. 🤷 I think the toolz logic ends up so much cleaner, but I'll accept this for now. 😂

magicgui/_magicgui.py
jni: Wowza.

magicgui/_magicgui.py
tlambert03: i agree.  looks much better that way.

magicgui/widgets/_function_gui.py
jni: Well, even so, if you're keeping _UNSET somewhere, can you explain why?

tlambert03: oh sorry, that sentinel is used for this line:
```
original_value = function.__globals__.get(func_name, _UNSET)
```
where we check to see what object the function name points to in the module in which it was defined.  obviously, it will _normally_ just be pointing to the function object itself, but if for some reason it is not, we want to reset it to that value when we're done. aaand, if it's not set at all (perhaps someone used `del`), then we use the `_UNSET` as a placeholder (instead of `None`)

magicgui/_magicgui.py
tlambert03: done

PR #118, opened 2021-01-22
More docs for main_window flag

HagaiHargil:
I've added a documentation entry for the main window feature as well as an example script.

I also introduced a very small modification for the code itself - the tooltip parser now removes backticks from the parameters' descriptions. 

I wasn't sure whether you want it based on current master or on #110, so I went with the latter. It also simplifies that small fix I just described, but I can obviously do whatever you like here.


examples/main_window.py
```diff
@@ -0,0 +1,39 @@
+import pathlib
+from enum import Enum
+
+from magicgui import magicgui
+
+
+class HotdogOptions(Enum):
+    """All hotdog possibilities"""
```
tlambert03, line 8: # 😂

tlambert03 (commented):
This is awesome @HagaiHargil, thank you so much!

I've merged the mainwindow PR, so can you merge master back into this PR and push again?

I've left a few nitpicky requests.  Much appreciated!

tlambert03:
oh one more thing... can you add the following line to `docs/_toc.yml`, right after "direct_api"?
```yml
    - file: usage/main_window
```

tlambert03 (approved):
fantastic.  Thanks again!

PR #140, opened 2021-01-31
Image widget

tlambert03:
basic support for an image widget. vendors the bare minimum from matplotlib to support:
- setting images from np.array (most 2D/3D forms), `PIL.Image.Image`, or file/url (requires `pillow`)
- setting clims with floats (vmin/vmax)
- setting norm with an mpl.colors.Normalize instance (though mpl doesn't need to be installed otherwise)
- setting cmap with the internal Colormap object or an mpl Colormap

currently does _not_ provide image rescaling (must be done before passing to image widget)

API is like this:

```python
from magicgui.widgets import Image

data = ... # data can be 2D or RGB(A) array, filepath or url, or PIL.Image.Image
image = Image()
image.value = data
# or 
image = Image(value=data)

# or to set more stuff
image.set_data(data, cmap=None, norm=None, vmin=None, vmax=None)
# where cmap can be a matplotlib cmap name if mpl is installed
# and norm can be a matplotlib norm if mpl is installed
# otherwise very basic ones are at `magicgui.types.Colormap` and `.Normalize`

# also supports individual setters like matplotlab Image:
image.set_clim
image.set_cmap
image.set_norm

```


GenevieveBuckley:
> currently does _not_ provide image rescaling (must be done before passing to image widget)

Is this a rescaling of image size, or intensity values for colourmaps, or both? Sorry if I skimmed too quickly and missed the answer.

tlambert03:
> Is this a rescaling of image size,

sorry, just rescaling image size is not provided (but even that is not a big deal, since most backends will be able to rescale).  Qt can do it, and browsers (ipywidgets) can do it too...  probably shouldn't even have mentioned it :joy:

tlambert03:
> > Is this a rescaling of image size,

PS. good to see you 😄

sofroniewn:
@tlambert03 can you provide a little bit more context on your vision for this part of the API. For example do you see it expanding to include other parts of the matplotlib API - for example histograms/ line plots etc through a `PlotWidget` or something similar?

Also do you see this as something that we might use inside napari or do you see this more as functionality to round out magicgui functionality independent of napari?

tlambert03:
> can you provide a little bit more context on your vision for this part of the API

I started wanting this during the alfa cohort meeting the other day, when someone showed a UI that had a bunch of buttons that were little images, and the user is supposed to click on the image that most resembles their sample.  Or, for @haesleinhuepf's main Assistant gui, which is also a grid of little images.  Basically, someone is going to want to have a little image somewhere in there :) so this allows it.

>  do you see it expanding to include other parts of the matplotlib API - for example histograms/ line plots etc through a PlotWidget or something similar?

not any time soon.  Just a little picture for now, that's all :)

> Also do you see this as something that we might use inside napari or do you see this more as functionality to round out magicgui functionality independent of napari?

I'd say it's just a vague step in a good direction there, but I'm thinking more of the first couple "gui builder" use cases than napari at the moment.

tlambert03:
> @haesleinhuepf's main Assistant gui, which is also a grid of little images. 

example: 
Untitled

sofroniewn:
> someone showed a UI that had a bunch of buttons that were little images, and the user is supposed to click on the image that most resembles their sample.

Ah ok, this is very helpful - yes I see that use case, it's present in `napari-animation` too, now that we capture keyframe thumbnails. That repo might be an interesting test case for a more advanced application of magicgui - not required, but could be fun to consider.

> For example do you see it expanding to include other parts of the matplotlib API - for example histograms/ line plots etc through a PlotWidget or something similar?

Just to add, I think this might be interesting to consider more in the context of our functional plugin ecosystem and autogenerating widgets that can display the results of analysis - but for this we might want to have a repo/ package that explicitly depends on matplotlib.

Let me know when you want a detailed review on this, I now have the necessary context - it looks exciting!

tlambert03:
> That repo might be an interesting test case for a more advanced application of magicgui - not required, but could be fun to consider.

definitely!  It's seems like just the right level of advanced to test this out.  Will try to have a look soon.

> I think this might be interesting to consider more in the context of our functional plugin ecosystem and autogenerating widgets that can display the results of analysis - but for this we might want to have a repo/ package that explicitly depends on matplotlib.

Yeah, would be very cool, and while I _can_ imagine how that fits in to the general concept of "mapping types to visual display" (e.g. `LinePlotData`), I think we'll need some very compelling use cases to warrant that level of feature creep any time soon :joy:

> Let me know when you want a detailed review on this, I now have the necessary context - it looks exciting!

i just finished up.  ready whenever.  I think `magicgui/widgets/_image.py` is probably the main thing you want to look at.

haesleinhuepf:
Hey @tlambert03 ,

awesome to see something like this in magicgui! I recently changed the assistant to exploit space better:
image

Putting the icons on top was [tricky](https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/master/napari_pyclesperanto_assistant/_gui/_Assistant.py#L98-L119).  Furthermore, I didn't mangage to have a floating layout. With a floating layout one could thake the dock-widget and use it as a toolbar underneath the menu bar. Does your implementation allow this by chance?

tlambert03:
While this PR simply adds the Image widget, all that stuff is possible, but not yet implemented:

> Putting the icons on top was tricky

That should be doable with the private [`_LabeledWidget`](https://github.com/napari/magicgui/blob/e9ac0917ceac69efd9b4038fc42e219464b5f09b/magicgui/widgets/_concrete.py#L522) class (currently, it _always_ puts the label to the left... but that is easy enough to expose publicly to put the label below the other widget as you did here)

> Furthermore, I didn't mangage to have a floating layout. 

That would also be great.  I'm currently working on extending the layouts to basic gridded layouts (from just HBox/VBox layouts)... and can probably add a floating layout after that.  (If you're anxious to take a crack at it now with Qt directly, here's an [example](https://doc.qt.io/qt-5/qtwidgets-layouts-flowlayout-example.html) of a custom Flow layout).

PR #160, opened 2021-02-23
Persist parameter values across sessions

### LINKED ISSUE #152: Add a cache keyword to read parameter values from disk
It could be nice if I could decorate a function with a 'cache' keyword so that when this function is run it will write to disk the last parameter values used to run it. Later, when it's re-run, these values will be read back and override any default keyword arguments that the function might have:

Example:

```python
@magicgui(cache=True, call_button='Run')
def test(a: int, b: int = 0) -> int:
    return a + b
```

On its first run, the values displayed will be 'empty' for `a` and `0` for `b`. If I run it with `a = 2` magicgui will write to the user's cache directory a snippet with `a = 2`, `b = 0`. This snippet will be looked-for and loaded during a consecutive call to `test`, as long as its run as a GUI (and not a standard function).

I tried implementing this behavior as a wrapper class but it became clear that I'll have to redo it for each and every magicgui script I have, and the work itself was quite tedious. If you think that this `cache` keyword fits the overall idea of magicgui I might try doing it myself, but not in the immediate future.

tlambert03:
closes #152.

This allows a user to save/restore the state of a `FunctionGui` by using the `persist=True` parameter either in `FunctionGui` or in the `@magicgui` decorator.  Parameter values are stored to disk (in `~/.config/magicgui/cache`) when they are changed, and reloaded upon future initialization (provided `persist` is still True the next time it is initialized):

```python
@magicgui(persist=True)
def func(x, y, z):
    ...
```

@HagaiHargil, give this a spin when you have a chance and let me know if anything needs to be changed.

### DIFF
--- a/magicgui/_magicgui.py
+++ b/magicgui/_magicgui.py
@@ -23,6 +23,7 @@ def magicgui(
     result_widget: bool = False,
     main_window: bool = False,
     app: AppRef = None,
+    persist: bool = False,
     **param_options: dict,
 ):
     """Return a :class:`FunctionGui` for ``function``.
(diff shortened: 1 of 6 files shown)


HagaiHargil:
I only now had a chance to use it and it works wonderfully! Thanks.

Wouldn't you prefer the cache to be in `/home/user/.cache`? Library like [appdirs](https://pypi.org/project/appdirs/) can take care of that, although pulling a dependency for this minor feature might be unwanted.

tlambert03:
yeah, i'd prefer to avoid the additional dependency, but I can certainly use the same destination folders as appdirs.  will do, thanks for trying it out

PR #166, opened 2021-02-28
fix range/slice edits

### LINKED ISSUE #162: RangeEdit and SliceEdit behave unexpectedly
**Describe the bug**
Control over the RangeEdit and SliceEdit widgets is currently sub-optimal, not producing the expected outputs in all cases. The same bug happens in both widgets, and is likely due to the same reason, so I'll simply focus on the RangeEdit object below for simplicity.

**To Reproduce**
I'd like to create a range object that starts and ends at the specified values:
```
from magicgui import magicgui


@magicgui(b={'start': 11, 'stop': 600})
def a(b: range):
    print("O_O")

a.show(run=True)
```
What I get, as seen in the screenshot, is 11, 99 and 1. The 99 value is sort-of the default value for high integers - I can increase it manually to 100 but it caps there, so I'm not really sure what's going on there. I tried also setting the `'max'` attribute to a high value, but that didn't change the outcome. As a side note I'd like to add that I never really understood that default limit for integer widgets to cap at 99\100...

No errors are reported in the console.

**Expected behavior**
I'd expect the three values to be 11, 600 and 1.

**Screenshots**
20210228_18h50m15s_grim

 - OS: Linux
 - backend: pyqt5
 - magicgui version [e.g. 0.1.0]: 0.2.6

tlambert03:
fixes #162,
exposes a `min`, `max` argument on Range/SlideEdit sliders that controls the min/max of the `start/stop/stop` sliders.  each accepts a scalar or a 3-tuple of scalars (for start/stop/step).

also made the default layout of a rangeEdit horizontal with labels.

### DIFF
--- a/magicgui/widgets/_bases/ranged_widget.py
+++ b/magicgui/widgets/_bases/ranged_widget.py
@@ -22,7 +22,7 @@ class RangedWidget(ValueWidget):
 
     _widget: _protocols.RangedWidgetProtocol
 
-    def __init__(self, min: float = 0, max: float = 100, step: float = 1, **kwargs):
+    def __init__(self, min: float = 0, max: float = 1000, step: float = 1, **kwargs):
         for key in ("maximum", "minimum"):
             if key in kwargs:
                 warn(
--- a/magicgui/widgets/_bases/ranged_widget.py
+++ b/magicgui/widgets/_bases/ranged_widget.py
@@ -40,6 +40,9 @@ class RangedWidget(ValueWidget):
         self.min = min
         self.max = max
         self.step = step
+        if kwargs.get("value") is not None:
+            # value may need to be reset *after* min max is set
+            self.value = kwargs["value"]
 
     @property
     def options(self) -> dict:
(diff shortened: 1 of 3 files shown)


tlambert03:
@HagaiHargil, the example in your original comment should now work:
```python
from magicgui import magicgui

@magicgui(b={'start': 11, 'stop': 600})
def a(b: range):
    ...

a.show(run=True)
```

but note also that you can just use the range object itself to set the default values:
```python
from magicgui import magicgui

@magicgui
def a(b=range(11, 600)):
    ...

a.show(run=True)
```

> As a side note I'd like to add that I never really understood that default limit for integer widgets to cap at 99\100...

I increased the default max to 1000 on the basic ranged value widgets... but that default is really so arbitrary that I'm not sure there's any reasonable default.  this really has the biggest implications for slider widgets, whose precision depends on the number of pixels per full range... so a really big default maximum is undesirable there.  In any case, for the `RangeEdit` spinboxes, I made the default huge since they probably will never be sliders... but you can still override with the `max`/`min` arguments to `RangeEdit` if you'd like

PR #170, opened 2021-03-01
Fail gracefully with persistence errors, better debounce

### LINKED ISSUE #169: Parameter value persistance causes error with napari layers as parameter types
**Describe the bug**
Using `@magicgui(persist=True)` works fine in a plugin, except the case when I have something like `image: napari.layers.Image` as an input parameter, e.g.  [here](https://github.com/brainglobe/cellfinder-napari/blob/6938485e38510a5688b3fccdefdf04eb2c322009/cellfinder_napari/cellfinder_napari.py#L31).

If I change the parameters, then re-start napari, the plugin doesn't start and I get an error. 

**To Reproduce**

`napari`
*Load plugin, change a parameter & close napari*
`NAPARI_CATCH_ERRORS=0 napari`
*Try to load plugin*
```
(napari-cellfinder) adam@garfield:~/projects/cellfinder-napari$ NAPARI_CATCH_ERRORS=0 napari
09:58:45 WARNING Could not import pyopencl. Falling back to scipy for CPU affine transforms
09:58:47 ERROR Unhandled exception:
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/napari/_qt/qt_main_window.py", line 450, in _add_widget
    self._add_plugin_dock_widget(key)
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/napari/_qt/qt_main_window.py", line 568, in _add_plugin_dock_widget
    wdg = Widget(**kwargs)
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/magicgui/_magicgui.py", line 200, in __call__
    widget = self.func(param_options=prm_options, **{**self.keywords, **kwargs})
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/magicgui/widgets/_function_gui.py", line 199, in __init__
    self._load(quiet=True)
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/magicgui/widgets/_function_gui.py", line 379, in _load
    super()._load(path or self._dump_path, quiet=quiet)
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/magicgui/widgets/_bases/container_widget.py", line 300, in _load
    getattr(self, key).value = val
  File "/home/adam/miniconda3/envs/napari-cellfinder/lib/python3.8/site-packages/magicgui/widgets/_bases/categorical_widget.py", line 38, in value
    raise ValueError(
ValueError: None is not a valid choice. must be in ()
```

This error doesn't occur if I:
* delete the cache file before re-starting napari
* Don't have a napari layer as parameter type in my decorated function

**Expected behavior**
All the parameters to be reloaded whether or not one of the inputs is a napari layer

 - OS: Ubuntu 18.04
 - backend: PySide2
 - magicgui version: 0.2.7

P.S. the parameter persistence is really helpful for my plugin, thanks!

tlambert03:
fixes #169 

Persistence is much tricker with complex objects like napari layers.  This PR makes it so that persistence fails gracefully.  We can consider how one might extend this persistence functionality to third-party libraries that register types (like napari) in a future PR.

This PR also fixes what @uschmidt83 observed about the rate limiting of saving to disk (previously, it was behaving more like a "throttle", which refuses to call the function if it's already been called within the last _n_ seconds.  Now it's a proper debounce... which will will postpone the call until _n_ seconds after the last invocation, so... if a slider moves, it won't save to disk for every intermediate event, but _will_ save with the last value.

### DIFF
--- a/magicgui/_util.py
+++ b/magicgui/_util.py
@@ -24,11 +24,11 @@ def debounce(function=None, wait: float = 0.2):
                 _store["last_call"] = time.time()
                 return fn(*_store["args"], **_store["kwargs"])
 
-            time_since_last_call = time.time() - _store["last_call"]
-            if time_since_last_call >= wait:
+            if not _store["last_call"]:
                 return call_it()
 
             if _store["timer"] is None:
+                time_since_last_call = time.time() - _store["last_call"]
                 _store["timer"] = Timer(wait - time_since_last_call, call_it)
                 _store["timer"].start()  # type: ignore
 
--- a/magicgui/_util.py
+++ b/magicgui/_util.py
@@ -1,12 +1,46 @@
 import os
 import sys
+import time
 from functools import wraps
 from pathlib import Path
-from time import time
 from typing import Optional
 
 
-def rate_limited(t):
+def debounce(function=None, wait: float = 0.2):
+    """Postpone function call until `wait` seconds since last invokation."""
+
+    def decorator(fn):
+        from threading import Timer
+
+        _store: dict = {"timer": None, "last_call": 0.0, "args": (), "kwargs": {}}
+
+        @wraps(fn)
+        def debounced(*args, **kwargs):
+            _store["args"] = args
+            _store["kwargs"] = kwargs
+
+            def call_it():
+                _store["timer"] = None
+                _store["last_call"] = time.time()
+                return fn(*_store["args"], **_store["kwargs"])
+
+            time_since_last_call = time.time() - _store["last_call"]
+            if time_since_last_call >= wait:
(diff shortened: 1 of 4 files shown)


jni (approved):

tlambert03:
Thanks for looking @jni

jni:
> Thanks for looking @jni

I thought it only fair after I threw you under the bus in #169 :joy:

Issue #173, opened 2021-03-02
connecting event vs value outputs


berl:
I found the `lambda e: callback(e.value.value)` [here](https://github.com/napari/magicgui/blob/46a968f69fbd852dcf07f76e822d21a80d794bbb/examples/file_dialog.py#L28) a little confusing, especially given the [current docs](https://napari.org/magicgui/examples/napari/napari_img_math.html) where the callback is shown receiving the `event` from the `called` signal and then pulling out the `event.value` to print it (or whatever).  

Idea: in addition to the lovely selection of signals, could `magicgui` provide a bare value returned by the decorated function? My feeling here is that someone using `magicgui` should have some way of defining what gets passed to a callback without having to make a `lambda` or debug how many times to get a `.value` attribute.  Side benefit is that downstream code used as callbacks wouldn't have to modify the inputs either.

tlambert03:
totally agree, that's a strange case! ... I honestly can't event remember now why you need a double `.value.value there 🤔 probably because the `FileEdit` is a compound widget (both a line edit and a button).  but that's no excuse, that should (and can) definitely just be `event.value`

> Idea: in addition to the lovely selection of signals, could magicgui provide a bare value returned by the decorated function? 


I agree, that that one additional level is kind of annoying, and we _could_ consider just calling callbacks with the new value (rather than the event object itself) ... but a fair amount of additional information about the event that triggered the change is unfortunately lost in that case.  I'm not saying it's not a worthwhile loss, but it doesn't come without a cost to flexibility.

> someone using magicgui should have some way of defining what gets passed to a callback without having to make a lambda or debug how many times to get a .value attribute.

outside of the strange `FileEdit` bug, it should should just _always_ be a single `.value`:

```python
def my_callback(event):
	new_widget_value = event.value

widget.changed.connect(my_callback)
```

as for the user indicating what the callback should get passed... that seems rather complicated actually.  It's pretty typical for a callback to be required to accept a standard argument. As mentioned above, we can definitely discuss & change _what_ that argument(s) should be... but I'm not sure it should be user-adjustable.

As for the lambda, that's purely optional syntax.  No users have to make lambdas if they don't want to, they can always use a named function as shown above.

So, thanks for opening this!  I do think it's worth considering changing what the callback is given.  That would be a relatively big breaking change though, so would require a major version change.  In the meantime, I'll fix the FileEdit and you can generally assume (with this one exception) that the new value is at `event.value`

berl:
[...]

PR #182, opened 2021-03-12
Allow callbacks to request event.value directly

tlambert03:
This is a proposal to address #173.  The event system (inherited from vispy, also used by napari) has the design that callbacks receive a single `Event` object, and must inspect that event object for things like `event.value` that they want to use.  Internally, that makes the `Event` object super flexible, but – as pointed out in #173 – for an end-user connecting callbacks, it probably almost always feels like an unnecessary layer.  This PR adds a backwards compatible, opt-in mechanism that would allow a callback to indicate that it wants `event.value` rather than `event`.

```python
from magicgui.events import EventEmitter

em = EventEmitter(type='test')
em.connect(print, callback_wants_value=True)
em(value=1)  # prints '1'
```

This behaves more like Qt Signals and Slots, where the slot/callback receives actual values, rather than some wrapper object.  One problem, of course, is that, internally, it _requires_ the event emitter to be called with `emitter(value=something)`.  Currently, _all_ events in magicgui follow that convention, so this PR doesn't go farther to impose any EventEmitter checking for compatibility with the `callback_wants_value` option.  (If a callback has requested value, and the even doesn't provide one, it currently just fails with a warning, rather than an exception).

@sofroniewn, @jni  ... since this is a concept that may well affect napari in the longer term (if we want to provide a more friendly callback-connecting API), curious to get your thoughts

### DIFF
--- a/magicgui/events.py
+++ b/magicgui/events.py
@@ -355,7 +355,7 @@ class EventEmitter:
     ):
         # connected callbacks
         self._callbacks: List[Union[Callback, CallbackRef]] = []
-        self._callback_wants_event: Dict[Union[Callback, CallbackRef], bool] = {}
+        self._callback_wants_value: Dict[Union[Callback, CallbackRef], bool] = {}
         # used when connecting new callbacks at specific positions
         self._callback_refs: List[Optional[str]] = []
 
@@ -451,7 +451,7 @@ class EventEmitter:
         position: Union[Literal["first"], Literal["last"]] = "first",
         before: Union[str, Callback, List[Union[str, Callback]], None] = None,
         after: Union[str, Callback, List[Union[str, Callback]], None] = None,
-        callback_wants_event: bool = True,
+        callback_wants_value: bool = False,
     ):
         """Connect this emitter to a new callback.
 
@@ -478,10 +478,10 @@ class EventEmitter:
         after : str | callback | list of str or callback | None
             List of callbacks that the current callback should follow.
             Can be None if no after-criteria should be used.
-        callback_wants_event : bool
-            If `True`, callback will receive a single `Event` argument.  If False
+        callback_wants_value : bool
+            If `False`, callback will receive a single `Event` argument.  If `True`
             callback will receive a single argument (the value at `event.value`).  By
-            default, `True`.
+            default, `False`.
 
         Notes
         -----
@@ -571,7 +571,7 @@ class EventEmitter:
 
         # actually add the callback
         self._callbacks.insert(idx, callback)
-        self._callback_wants_event[callback] = callback_wants_event
+        self._callback_wants_value[callback] = callback_wants_value
         self._callback_refs.insert(idx, _ref)
         return callback  # allows connect to be used as a decorator
 
@@ -645,7 +645,7 @@ class EventEmitter:
 
             rem: List[CallbackRef] = []
             for cb in self._callbacks[:]:
-                wants_event = self._callback_wants_event[cb]
+                wants_value = self._callback_wants_value[cb]
                 if isinstance(cb, tuple):
                     obj = cb[0]()
                     if obj is None:
@@ -660,7 +660,7 @@ class EventEmitter:
                     self._block_counter.update([cb])
                     continue
 
-                self._invoke_callback(cb, event, wants_event)
+                self._invoke_callback(cb, event, wants_value)
                 if event.blocked:
                     break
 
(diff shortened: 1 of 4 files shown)


sofroniewn:
At a high level I love it - the discussion in #173 really resonated, and I'd love to see us maybe even do more of this within napari too - or at least as the napari to external world interface. Within napari once we've finished the evented_model refactor all our events will use `.value` as well and only a couple in the `evented_list`/ `evented_set` will use multiple things. I think the simplicity gained for plugin developers would be large there too. We don't have to worry about that now in napari, but just broadly giving my support to this effort.

As to the details, this PR seems like the best minimal change that enables opt-in flexibility.

I could have even gone the other way, which would have been opt out `em.connect(print, callback_wants_events=True)`, though that would be an immediate breaking change. (You could do something fancy like put a deprecation note in the `event.value` getter like (this should no longer be used, if you only want the value get it here ...).

Longer term, I'd even be in favor of a change to the event emitter too which was `EventEmitter(emits_values: bool = True)`, though once we set that on napari/ magicgui side it's then never possible to get events out, unless you had also `em.connect(print, callback_wants_events=True)` that would over rule it?

Anyway, starting out opt-in is probably best and it would be great to bring this to napari, and we could start using it slowly internally in an opt-in fashion too.

If we think back to our "event handler" discussion of last year, we decided that the "evented model" was the first phase of improvements, this "connect with value" could now be the second phase! 🚀

tlambert03:
Thanks for taking the time to read/weigh in!  glad you're generally on board.  I agree it makes for a much nicer UX.

> I could have even gone the other way, which would have been opt out `em.connect(print, callback_wants_events=True)`, though that would be an immediate breaking change.

I agree the other way should be the default, but as that will be very breaking, i'd like to leave that change for 0.3.0.  I can add a `FutureWarning` now though for a couple versions.

> I'd even be in favor of a change to the event emitter too which was EventEmitter(emits_values: bool = True), though once we set that on napari/ magicgui side it's then never possible to get events out

yes! This would be great to give some up-front validation.  I think you would still be able to get events out with `callback_wants_value=False`...

jni:
I'm on board with the idea, I think we should support both long-term with particular syntax, but I don't love kwarg names mooted so far (`callback_wants`). It seems verbose and... I dunno, strange?

Now, if I read the argument above, this works with `em(value=1)`, but someone could do `em(value=1, previous=0)` and things will break, right? Since this is another thing we are considering in the future, what about:

`event.connect(callback, what=['value'])`, vs the (admittedly magical) `what=event` which would be the default now, vs `what=['value', 'prev']` which would connect *two* arguments. You could even do `what=None` for things that just want to know that something happened but never actually look at the event. And you could map keyword arguments to event values, `what={'callback_kwarg1': 'value'}`.

Does any of the above make sense? I'm not much of an event-driven programming person, but maybe that's a good thing here. 😅

tlambert03:
> event.connect(callback, what=['value']), vs the (admittedly magical) what=event which would be the default now, vs what=['value', 'prev'] which would connect two arguments. You could even do what=None for things that just want to know that something happened but never actually look at the event. And you could map keyword arguments to event values, what={'callback_kwarg1': 'value'}.

love this!  definitely better and more flexible than the `callback_wants` stuff.  in the longer term, my primary hesitation with this whole thing at this point is the inability to validate this stuff at the moment of callback connection.  In Qt, if you try to connect to a signal with a callback that has an incompatible signature, you immediately get an error.  That's kind of critical... but could come later.  Anyway, i'll incorporate what you've suggested here and go from there

pranathivemuri:
Hi @tlambert03 I am trying to figure out how if a table widget's value is changed on napari then that changed value can trigger an event and the value can be recorded and a shapes layer on napari that reflects property of that bounding box can be changed based on the changed even in relation to this issue here - https://github.com/napari/napari/issues/2427 

Seems like this is something I would need to achieve the above, so wondering if this PR is relevant and if it is ready to be merged to master?

tlambert03:
> Seems like this is something I would need to achieve the above

Hi @pranathivemuri... not quite: this PR isn't adding any new functionality, it's simply changing slightly the value your callback would receive when you use `widget.changed.connect(my_callback)`.  

> if a table widget's value is changed

if you mean you're creating a `magicgui.widgets.Table` and want something to happen when the user changes a value in the table, I'm afraid that isn't implemented yet for individual cells in a table (... sorry, I can definitely see how that would be necessary for something like napari/napari#2427).  Table's are a special case for the `widget.changed` callback, and I stopped short of implementing it just yet (see note here) since I didn't want to just blinding emit the entire table each time any cell was changed.  But thanks for bringing it up again!  I've made a new issue  for this (#205) and will try to get to it soon.

pranathivemuri:
@tlambert03 That is what I meant, thanks for the quick response!

PR #183, opened 2021-03-12
add `RadioButtons` widget

tlambert03:
This adds `widgets.RadioButtons` which has the same API as `widgets.ComboBox` (with the addition of an `orientation` parameter).  Currently, only exclusive radio groups are supported. 

shown here with horizontal orientation in a vertical layout:

Untitled


sofroniewn (approved):
Looks good to me! By default will you get the opposite orientation of the orientation of the main layout if possible? So always some combo of vertical + horizontal. Or is that maybe too magical? You could always fallback on a single default if nothing else makes sense

tlambert03:
> By default will you get the opposite orientation of the orientation of the main layout if possible?

I was just gonna stick with vertical by default... _someone_ won't like the magic :)

sofroniewn:
> I was just gonna stick with vertical by default... someone won't like the magic :)

fine with me!

jni (approved):
:heart:

PR #199, opened 2021-03-24
Compute correct widget width for rich text labels

maweigert:
Currently, the width of label widgets that contain rich text (such as html tags) is computed based on the raw text alone. This may result in excessively large widgets, e.g. when including images.

For instance the following 

```python

#download test image 
!wget https://raw.githubusercontent.com/napari/magicgui/master/resources/hotdog.jpg -O very_impressive_flashy_logo_for_my_awesome_napari_plugin.jpg


from magicgui import magicgui

@magicgui(
    label=dict(widget_type='Label', label='<img src="very_impressive_flashy_logo_for_my_awesome_napari_plugin.jpg" width="64">Text'))
def widget(label):
    pass

widget.show()

```

leads to 

Screenshot 2021-03-24 at 16 38 07

The PR fixes how the default size of the label widget is computed for rich text labels. The above code then yields

Screenshot 2021-03-24 at 16 38 16


tlambert03 (approved):
awesome!  thanks!

tlambert03:
bummer ... `mightBeRichText ` doesn't look to be in the `Qt` namespace in pyside2... we'll have to figure out what's going on there.

tlambert03:
from the [qt source code](https://github.com/qt/qtbase/blob/376e3bd8ecf40881685714f6f19e12d68e92127e/src/gui/text/qtextdocument.cpp#L85-L96), looks like all `mightBeRichText` isn't doing much: 

>   This function uses a fast and therefore simple heuristic. It
    mainly checks whether there is something that looks like a tag
    before the first line break. Although the result may be correct
    for common cases, there is no guarantee.

so we can just use regex and do:
```python
def might_be_rich_text(text):
    import re
    return bool(re.search("<[^\n]+>", text))
```

maweigert:
> bummer ... `mightBeRichText ` doesn't look to be in the `Qt` namespace in pyside2... we'll have to figure out what's going on there.

Mhm, seems `mightBeRichText` lives in different submodules for both backends:

```python
PySide2.QtGui.Qt.mightBeRichText
PyQt5.QtCore.Qt.mightBeRichText
```

tlambert03:
thanks!

PR #200, opened 2021-03-24
Add cache location hint to persist option documentation

maweigert:
Small change in the doc of the `persist` option (the earlier given cache location is different on OSX for example)

### DIFF
--- a/magicgui/_magicgui.py
+++ b/magicgui/_magicgui.py
@@ -58,8 +58,9 @@ def magicgui(
         A backend to use, by default ``None`` (use the default backend.)
     persist : bool, optional
         If `True`, when parameter values change in the widget, they will be stored to
-        disk (in `~/.config/magicgui/cache`) and restored when the widget is loaded
-        again with ``persist = True``.  By default False.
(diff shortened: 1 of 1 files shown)


tlambert03:
thanks again.  Since this (and pretty much all of the options on `@magicgui`) are just parameters for `magicgui.widgets.FunctionGui`.  We should also add it to the docs there too.  Also, I don't love adding a private module.parameter to the docstring like that, so maybe we should make `_util.py` public?  Alternatively, fyi, you can get the specific cache path of each generated widget at `widget._dump_path`.  Not sure if you want to mention that...  we could make that public (and possible change the name to `cache_path`).  Thoughts?

maweigert:
Sure, maybe even add that cache location may vary is already enough (instead of changing visibility of a whole module). I was just a bit tripped up by the current doc and that might be the case for others to.

tlambert03:
> I was just a bit tripped up by the current doc and that might be the case for others to.

yeah, def... that was outdated.  much appreciated.  OK, let's just use this as is for now and I may extend it later.  thanks again!

tlambert03 (approved):

PR #201, opened 2021-03-25
fix enum choices in radiobuttons

tlambert03:
this fixes a bug that @jni encountered wherein `RadioButtons` used with Enums weren't being populated with choices


jni (approved):

jni:
Works great with affinder! :blush:

PR #211, opened 2021-04-05
Display current slider value (editable)

### LINKED ISSUE #19: Display slider current value and range min/max in the magicgui widget
I've been doing some work involving magicgui, and have noticed that sliders don't show any indication of what value currently is selected.

For a minimal working example, in [napari_parameter_sweep.py](https://magicgui.readthedocs.io/en/latest/examples/napari_parameter_sweep/) the image updates but the slider doesn't show any indication of what the current selected value is, or for that matter the min & max of the slider range (you'd have to go digging in the code for that). This is less than ideal for a parameter sweep, where you want to be able to make a note of what the best values were at the end.

tlambert03:
closes #19 with a 2-way editable indicator of slider value.  

https://user-images.githubusercontent.com/1609449/113525891-86d18880-9585-11eb-85b2-5c821e05d44c.mov

This PR also refactors the way float sliders are done (puts more of it in the backend... meaning other backends will need to reimplement float sliders, but doing it in a backend agnostic way was too ugly).

#19 also asked for min/max range to show, but I think I'll leave that for another PR.


sofroniewn (approved):
LGTM. I think "readout" is a fine kwarg. Other options might have been "text", but maybe that's ambiguous.

Something cool one day might be to add an additional layout option too like "condensed" or "hidden" (not sure the right words here) and then to make just the text box visible by default and the slider appear when the arrow is clicked like in this movie from photoshop.

The nice thing about that is that it is very space efficient if you have lots of parameters like this. Not sure how hard it would be to do in Qt!

https://user-images.githubusercontent.com/6531703/113594587-bff11380-95ec-11eb-9123-11f2404cd4a9.mov

tlambert03:
>  I think "readout" is a fine kwarg.

Yeah, I had a hard time picking a kwarg, so I just took the one from [ipywidgets](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#IntSlider) 😂  

> Something cool one day might be to add an additional layout option too like "condensed" or "hidden" (not sure the right words here) and then to make just the text box visible by default and the slider appear when the arrow is clicked like in this movie from photoshop.

indeed!  that's lovely.  could do someday.

schoinh:
Thanks, I'm excited to try this out! Hoping it'll be easy to plug in magicgui into our otherwise all-Qt plugin 🤞🏻

tlambert03:
hope so too.  as always, let me know if you run into hurdles :)

Issue #145, opened 2021-02-02
Optional types should map to the same widgets as non-optional and be bound to the default None


jni:
**Describe the bug**

Annotating a function argument with `pathlib.Path` gives me a native file selection widget, which is 👍! Changing it to `Optional[pathlib.Path] = None` (a) fails to find the file widget anymore, and (b) gives an unbound argument error when run:

```pytb
/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/widgets/_bases/widget.py:66: FutureWarning:

EmptyWidget.__init__() got unexpected keyword arguments {'mode'}.
In the future this will raise an exception

  warnings.warn(
Traceback (most recent call last):
  File "/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/widgets/_bases/value_widget.py", line 44, in <lambda>
    lambda *x: self.changed(value=x[0] if x else None)
  File "/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/events.py", line 603, in __call__
    self._invoke_callback(cb, event)
  File "/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/events.py", line 621, in _invoke_callback
    _handle_exception(
  File "/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/events.py", line 619, in _invoke_callback
    cb(event)
  File "/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/widgets/_function_gui.py", line 179, in <lambda>
    self._call_button.changed.connect(lambda e: self.__call__())
  File "/Users/jni/conda/envs/all2/lib/python3.9/site-packages/magicgui/widgets/_function_gui.py", line 252, in __call__
    raise TypeError(msg) from None
TypeError: missing a required argument: 'fn' in call to 'widget(fn: Optional[pathlib.Path])'.
To avoid this error, you can bind a value or callback to the parameter:

    widget.fn.bind(value)

Or use the 'bind' option in the magicgui decorator:

    @magicgui(fn={'bind': value})
    def widget(fn: Optional[pathlib.Path]): ...
```

**To Reproduce**

```python
from typing import Optional
import pathlib
from magicgui import magicgui


@magicgui(fn={'mode': 'r'}, call_button='Run')
def widget(fn: Optional[pathlib.Path] = None):
    print(fn)


widget.show(run=True)
```

**Expected behavior**
[...]

PR #227, opened 2021-04-30
Add ValueWidget._nullable, and enable for categorical widgets

tlambert03:
This is a partial solution to #145.  It introduces the concept of a "nullable" to `ValueWidget` (though it's not yet really doing anything there).  It does do something in categorical widget where, if the widget is nullable, a new option `----` will appear in the combobox, which returns `None` (both of those are adjustable via `widget.null_string`, and `widget.null_value`).

In `type_map.py`, we look for an annotation of `Optional[something]` and then set `nullable=True` by default.


tlambert03:
added optional `Path` as well

PR #249, opened 2021-07-11
Specify XDG_RUNTIME_DIR to avoid warnings showing up in the docs

hmaarrfk:
I find 
image

Sprinkled in the docs 
https://napari.org/magicgui/usage/direct_api.html

I wonder if this will address it.

Can we monitor the temporary output?

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -43,6 +43,7 @@ jobs:
         run: jupyter-book build docs
         env:
           DISPLAY: ":99.0"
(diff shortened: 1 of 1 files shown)


tlambert03:
thanks @hmaarrfk!

> Can we monitor the temporary output?

probably ... but I think it would probably take more work to figure that out 😂 , so lets just give it a try! and then iterate if if doesn't work

tlambert03:
hooray! https://napari.org/magicgui/usage/direct_api.html

hmaarrfk:
Cool great!

PR #263, opened 2021-07-28
Fix missing event emission when nullable widget is set to null value.

### LINKED ISSUE #260: No event when selection of --- (None) when using Optional annotation
**Describe the bug**

When using `Optional` annotation for ComboBox then change on `--` value does not fire `changed` event when selecting of value cause such event

**To Reproduce**

Here example of code, base on the base `Layer` registration code.   

```python
from typing import Optional

from magicgui import register_type
from magicgui.widgets import create_widget, Label, Container
from qtpy.QtWidgets import QApplication

def get_choices(gui):
    return ["aa", "bb", "cc"]

def get_callback(gui, result, return_type):
    print(result)
    return 1


def check_connect(event):
    label.value = choice.current_choice
    print(event)


class ExampleClass:
    pass

register_type(ExampleClass, choices=get_choices, return_callback=get_callback)



app = QApplication([])
choice = create_widget(annotation=Optional[ExampleClass])
choice.changed.connect(check_connect)
label = Label()
Container(widgets=(choice, label)).show()

app.exec_()
```

**Expected behavior**

The event should be emitted.

**Screenshots**

 - OS: Linux
 - backend: Qt 5.15.4
 - magicgui version 0.2.10

tlambert03:
fixes #260

### DIFF
--- a/magicgui/backends/_qtpy/widgets.py
+++ b/magicgui/backends/_qtpy/widgets.py
@@ -626,9 +626,7 @@ class ComboBox(QBaseValueWidget, _protocols.CategoricalWidgetProtocol):
         self._qwidget.currentIndexChanged.connect(self._emit_data)
 
     def _emit_data(self, index: int):
-        data = self._qwidget.itemData(index)
-        if data is not None:
-            self._event_filter.valueChanged.emit(data)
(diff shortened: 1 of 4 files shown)


Czaki:
It works

PR #276, opened 2021-08-21
Fix call button text and and param options when using decorator on class method

### LINKED ISSUE #275: no call button label for magicgui on bound method
**Describe the bug**
Not sure if this is a bug or just out of normal use case... the call button label is not appearing correctly when magicgui decorator is used on a bound method rather than a function

**To Reproduce**
The following in ipython with %gui qt shows a button with 'run' rather than 'my_func'

```python
ifrom magicgui import magicgui

class MyObject:
    @magicgui(call_button='my_func')
    def my_func(self):
        print('something')

my_object = MyObject()

my_object.my_func.show()

```

**Expected behavior**
Call button should be called 'my_func'

**Screenshots**
image

 - OS: mac
 - magicgui version 0.2.10

tlambert03:
fixes  #275

### DIFF
--- a/magicgui/widgets/_function_gui.py
+++ b/magicgui/widgets/_function_gui.py
@@ -316,7 +316,7 @@ class FunctionGui(Container, Generic[_R]):
         """Return a copy of this FunctionGui."""
         return FunctionGui(
             function=self._function,
-            call_button=bool(self._call_button),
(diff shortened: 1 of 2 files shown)


alisterburt (approved):
Definitely fixes the bug, awesome! Thanks :)

PR #301, opened 2021-10-17
Table widget updates

hanjinliu:
This PR adds three changes in `Table` widget:

1. **Float is formatted before displayed**. Four digits are displayed for now, which seems enough for most purposes, but I think we should discuss more if there are any better solutions (such as `"enable_exp"` property that allow exponential notation like "5.2e-02", or dynamically change float format upon resized).
2. **Header is resizable**.
3. New way of **changing editability of items**. The `setFlags` method is useful when we have to set different editability flags separately for each item. Here we only have to change editability of table itself, `setEditTriggers` is faster because we don't have to run for loops.

### DIFF
--- a/magicgui/backends/_qtpy/widgets.py
+++ b/magicgui/backends/_qtpy/widgets.py
@@ -1206,7 +1206,7 @@ class _ItemDelegate(QtW.QStyledItemDelegate):
             value = int(text)
         except ValueError:
             try:
-                value = float(text) # type: ignore
+                value = float(text)  # type: ignore
             except ValueError:
                 pass
 


hanjinliu:
Thank you for your check!
It seems like there is no simple way to get the displayed text from a table widget.
And yes, I'm aware of the concept of non-backend-dependency. I'll have to try making a consensus between different backends in the future.

tlambert03:
sure that test is fine.  (probably would have made it a pure function that didn't require a qapp and instantiation of the item delegate... but this is fine too, just want to make sure we're at least roughly testing new things).

> Sorry for pushing many times. I was not familiar with the pre-commit system...

no worries!  to avoid that in the future, you can install pre-commit locally with `pre-commit install` (read more [here](https://pre-commit.com/#quick-start))

PR #302, opened 2021-10-21
Add stacklevel to provide better information about place of problem

Czaki:
Currently magicgui warnings point to place in magicgui code. After this change it will point to place from where function/constructor is called.


tlambert03:
thanks, much better.  can release again with this if you think i should?

Czaki:
I think yes. It will simplify plugin update process.

PR #303, opened 2021-10-22
Fix unbound variable in format_number

tlambert03:
looks like we forgot the case of strings in tables in #301 ... if a string can't be cast to either an int or float, we now have an unbound variable.  This fixes that and adds a test
cc @hanjinliu


hanjinliu:
Oh no, I'm sorry I was careless!

PR #316, opened 2021-11-11
Support partials and tz.curry

### LINKED ISSUE #315: Add support for partial/curried functions
This is standard magicgui and works fine:

```python
from magicgui import magicgui

def add(a: int, b: int) -> int:
    return a + b

gui = magicgui(add, result_widget=True, auto_call=True)
gui.show(run=True)
```

However, if I want to partially run a function before passing it to magicgui, I would normally use `functools.partial` or `toolz.curry`. The first one fails:

```python
from magicgui import magicgui
from functools import partial

def add(a: int, b: int) -> int:
    return a + b

gui = magicgui(functools.partial(add, 5), result_widget=True, auto_call=True)
gui.show(run=True)
```

with:

```pytb
  File "/home/jni/projects/play/magicgui-partial.py", line 7, in <module>
    gui = magicgui(partial(add, 5), result_widget=True, auto_call=True)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/_magicgui.py", line 90, in magicgui
    return _magicgui(**locals())
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/_magicgui.py", line 246, in _magicgui
    return inner_func(function)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/_magicgui.py", line 241, in inner_func
    return magic_class(func, **kwargs)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/widgets/_function_gui.py", line 156, in __init__
    or f"{function.__module__}.{function.__class__}"
AttributeError: 'functools.partial' object has no attribute '__module__'
```

And the second one:

```python
from magicgui import magicgui
import toolz as tz

@tz.curry
def add(a: int, b: int) -> int:
    return a + b

gui = magicgui(add(5), result_widget=True, auto_call=True)
gui.show(run=True)
```

fails with:

```pytb
  File "/home/jni/projects/play/magicgui-partial.py", line 8, in <module>
    gui = magicgui(add(5), result_widget=True, auto_call=True)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/_magicgui.py", line 90, in magicgui
    return _magicgui(**locals())
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/_magicgui.py", line 246, in _magicgui
    return inner_func(function)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/_magicgui.py", line 241, in inner_func
    return magic_class(func, **kwargs)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/widgets/_function_gui.py", line 162, in __init__
    widgets=list(sig.widgets(app).values()),
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/signature.py", line 222, in widgets
    {n: p.to_widget(app) for n, p in self.parameters.items()}
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/signature.py", line 222, in <dictcomp>
    {n: p.to_widget(app) for n, p in self.parameters.items()}
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/signature.py", line 139, in to_widget
    widget = create_widget(
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/widgets/_bases/create_widget.py", line 91, in create_widget
    widget = wdg_class(**kwargs)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/widgets/_concrete.py", line 164, in __init__
    super(cls, self).__init__(**kwargs)
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/widgets/_bases/ranged_widget.py", line 45, in __init__
    self.value = val
  File "/home/jni/miniconda3/envs/all/lib/python3.9/site-packages/magicgui/widgets/_bases/ranged_widget.py", line 57, in value
    if not (self.min <= float(value) <= self.max):
ValueError: could not convert string to float: '__no__default__'
```

tlambert03:
This implements support for functions wrapped in functools.partial. closes #315

There is a slight "catch" ... which is really due to the way that partial handles signatures when providing positional vs keyword arguments.  Take this example

```py
    def some_func(x: int, y: str) -> str:
        return y + str(x)

    wdg1 = magicgui(partial(some_func, 1))  # first position
    wdg2 = magicgui(partial(some_func, y='hi'))  # second kwarg
```

wdg1 will look like this:

Untitled
and wdg2 will look like this

Untitled2

they both _behave_ like the corresponding partials.  but wdg2 shows the widget for `y`, whereas `wdg` doesn't show `x`

To avoid that behavior automatically, we'd need to do a little special casing.  It's possible, but prefer to leave to another PR

### DIFF
--- a/magicgui/signature.py
+++ b/magicgui/signature.py
@@ -26,6 +26,8 @@ if TYPE_CHECKING:
     from magicgui.widgets import Container
     from magicgui.widgets._bases import Widget
 
+TZ_EMPTY = "__no__default__"
+
 
 def make_annotated(annotation=Any, options: dict = None) -> _AnnotatedAlias:
     """Merge a annotation and an options dict into an Annotated type.
@@ -134,7 +136,7 @@ class MagicParameter(inspect.Parameter):
         from magicgui.widgets._bases import create_widget
         from magicgui.widgets._bases.value_widget import UNSET
 
-        value = UNSET if self.default is self.empty else self.default
+        value = UNSET if self.default in (self.empty, TZ_EMPTY) else self.default
         annotation, options = split_annotated_type(self.annotation)
         widget = create_widget(
             name=self.name,
--- a/magicgui/widgets/_function_gui.py
+++ b/magicgui/widgets/_function_gui.py
@@ -155,10 +155,16 @@ class FunctionGui(Container, Generic[_R]):
         # access attributes (like `__name__` that only function objects have).
         # Mypy doesn't seem catch this at this point:
         # https://github.com/python/mypy/issues/9934
-        self._callable_name = (
-            getattr(function, "__name__", None)
-            or f"{function.__module__}.{function.__class__}"
-        )
+        name = getattr(function, "__name__", None)
+        if not name:
+            try:
+                name = f"{function.__module__}.{function.__class__}"
+            except AttributeError:
+                # partials
+                f = getattr(function, "func", None)
+                name = getattr(f, "__name__", None) or str(function)
+        self._callable_name = name
+
         super().__init__(
             layout=layout,
             labels=labels,
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,6 +70,7 @@ testing =
     %(tqdm)s
     %(image)s
     matplotlib
+    toolz
 dev =
     ipython
     black
--- a/tests/test_magicgui.py
+++ b/tests/test_magicgui.py
@@ -761,7 +761,7 @@ def test_curry():
     assert isinstance(wdg.y, widgets.LineEdit)
     assert not hasattr(wdg, "x")
     assert wdg("sdf") == "sdf1"
-    assert wdg._callable_name == "some_func"
+    assert wdg._callable_name == "some_func2"
 
     wdg2 = magicgui(some_func2(y="sdf"))
     assert len(wdg2) == 3  # keyword arguments don't change the partial signature
--- a/tests/test_magicgui.py
+++ b/tests/test_magicgui.py
@@ -747,3 +747,25 @@ def test_partial():
     assert isinstance(wdg.y, widgets.LineEdit)
     assert wdg2.y.value == "sdf"
     assert wdg2(1) == "sdf1"
+
+
+def test_curry():
+    import toolz as tz
+
+    @tz.curry
+    def some_func2(x: int, y: str) -> str:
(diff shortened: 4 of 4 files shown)


tlambert03:
added support for `tz.curry` too... the main difference there was that toolz uses the string `__no__default__ ` as a sentinel, instead of `Parameter.empty`, to indicate the lack of a value.

```py
In [1]: from functools import partial

In [2]: from toolz import curry

In [4]: def f(x: int, y: str): ...

In [5]: a = partial(f, y='hi')

In [6]: b = curry(f)(y='hi')

In [7]: import inspect

In [8]: inspect.signature(a)
Out[8]: <Signature (x: int, *, y: str = 'hi')>

In [9]: inspect.signature(b)
Out[9]: <Signature (x: int = '__no__default__', *, y: str = 'hi')>
```

so we need to special case that here.  @jni if you wanted to upstream anything, it would be to see whether they can use `Parameter.empty` as a sentinel instead...

jni:
@tlambert03 where does parameter.empty come from exactly?

jni:
(Also this is incredible, I thought my issue might eventually be fixed in a few months... 😂)

tlambert03:
> @tlambert03 where does parameter.empty come from exactly?

```py
In [1]: import inspect

In [2]: inspect.Parameter.empty
Out[2]: <class 'inspect._empty'>
```

tlambert03:
and...

```py
In [3]: def f(x): ...

In [4]: inspect.signature(f).parameters
Out[4]: mappingproxy({'x': <Parameter "x">})

In [5]: inspect.signature(f).parameters['x'].default
Out[5]: <class 'inspect._empty'>
```

tlambert03:
> (Also this is incredible, I thought my issue might eventually be fixed in a few months... 😂)

turned out to be relatively trivial!

tlambert03:
> (Also this is incredible, I thought my issue might eventually be fixed in a few months... 😂)

can I assume this is approved? wanna hit the button? :)

Issue #98, opened 2021-01-14
Support pydantic types


tlambert03:
As @sofroniewn suggested, pydantic's constrained types make for a natural way to express things like ranges and widget parameters as type annotations.  We should detect and support pydantic types

aeisenbarth:
Currently I am building my user interface using the direct widget API and noticed I should avoid repetition by keeping widget constraints (type, min/max) in a central place, my data model which is a Pydantic class and has all these constraints. As a Pydantic class (from which I can generate documentation) it also can hold widget texts (Pydantic `title`) and tooltips (Pydantic `description`). Basically I ended up with rather repetitive code that feeds each Pydantic field property to an argument of the widget constructor.

It would be nice to generate a magicgui either from a pydantic class (like magic-class from basic dataclasses), or to have something similar to `widgets.create_widget()` with a Pydantic field as single required argument.

tlambert03:
thanks @aeisenbarth ... I have a local branch where I started to implement this.  will try to get back on it oon

imagejan:
Seeing the discussions in #474 and #475, is the plan still to have first-class support for Pydantic classes?
Or should I rather look into using `@dataclass` or `@guiclass`?

imagejan:
> Can you read through this topic on imagesc and let me know what your specific desired use case would be?
> [forum.image.sc/t/building-a-napari-widget-from-a-pydantic-model/90257/5](https://forum.image.sc/t/building-a-napari-widget-from-a-pydantic-model/90257/5)

Thanks @tlambert03, that's a very useful forum post!
I opened a new topic ([Using magicgui to create pydantic class instances](https://forum.image.sc/t/using-magicgui-to-create-pydantic-class-instances/92520)) to describe my intended use case and post a few more questions 🙂.

PR #318, opened 2021-11-14
Initial support for pydantic models

tlambert03:
A start on #98: support for pydantic models.

I say a "start" because I'm certain there are going to be types that are rather easy to declare in pydantic, for which generating a widget is not straightforward.

@aeisenbarth, I'd be curious to hear your experience if you pull this and try it with your models.  See examples/pydantic_model ... (basic idea is that `magicui()` now also takes a BaseModel as an argument


sofroniewn (commented):
Increadible, I love that this is pretty much a straight up addition on top of existing magicgui code and didn't require complex changes to existing code

> I say a "start" because I'm certain there are going to be types that are rather easy to declare in pydantic, for which generating a widget is not straightforward.

Indeed. I could imagine some types might never be possible, or take much much longer and so probably fine to have some warnings - "type not yet supported" etc.

I was looking through `pydantic_type` and `model_field_widget` but it wasn't quite clear to me exactly how to tell which types were being supported yet. I know PR is still fresh, but might be nice to make it clear somewhere in code where the supported types were "listed" so it's easier to tell what is/ isn't supported. But I should say I only took a quick look at code, so might have just missed something

Also will this now make supporting jsonschema -> widgets much easier as we can do jsonschema -> pydantic -> widgets? If so I think that would make a great example/ call out because I see that as a major usecase of this too. Maybe once this is merged we could then drop our vendored qtjsonschema and use this instead (cc @ppwadhwa @goanpeca).

All in all though awesome stuff! Happy to do a more detailed review too, these were just some initial thoughts

aeisenbarth:
Thanks, this looks good! In my GUI, I wasn't annotating a function signature but I subclassed Container and added widgets using the direct widget API. This GUI class is then registered as a dock, and it has a `model` attribute which is the Pydantic model (the GUI class itself was not a Pydantic class). So it took me a while to see how I best integrate this.

I tried two approaches:
1. Keep my Pydantic model under the `model` attribute, convert it to a widget and add it to the container
    ```
    class MyDock(Container):
        …
        inputs = model_widget(self.model.__class__) # Important: I used to instantiate my model, but here the class is needed.
        self.append(inputs)
    ```
    - Pro: It is easy to add more widgets (buttons, layer selector) and you are not required to have them in the model.
    - Issues: I tried the `Tuple[int, int]` annotation and it can only have a label/tooltip on container of the tuple, not different texts individually on the inputs for each int. Alternatively I used a nested pydantic model (`class Shape(BaseModel)  … grid_shape: Shape = Field(title="Shape", ui_layout="horizontal")`) with two int attributes, but the container created from this submodel would not take up the submodel's properties ("grid_shape" instead of "Shape", and vertical layout instead of horizontal)

2. Turn the dock into a Pydantic model (`class MyDock(BaseModel`)
    - In my UI init function added to this model class, it is not clear how to access the input widgets to connect them to further functions because accessing an attribute gives its value, not the created widget.
    - Also I could not append extra widgets that are not represented in the model. Underscored attributes (`_my_private_attribute`) would cause errors in Pydantic.

goanpeca:
>pydantic -> widgets? If so I think that would make a great example/ call out because I see that as a major usecase of this too. Maybe once this is merged we could then drop our vendored qtjsonschema and use this instead (cc @ppwadhwa @goanpeca).

Thanks for the ping @sofroniewn. We could do `pydantic -> widgets` directly, but need to make sure to add the ability to provide custom widgets that follow an interface, we could actually reuse some of the qtjsonschema code.

tlambert03:
closing this in favor of #446

PR #322, opened 2021-11-17
adds waveform generator exemple

glyg:
Adds a waveform app example, with an embedded matplotlib plot

### DIFF
--- a/examples/waveform.py
+++ b/examples/waveform.py
@@ -7,7 +7,20 @@ import numpy as np
 from matplotlib.backends.backend_qt5agg import FigureCanvas
 from scipy import signal
 
-from magicgui import magicgui, widgets
+from typing_extensions import Annotated
+
+
+from magicgui import magicgui
+
+from magicgui import magicgui, type_map, widgets, widgets, register_type
+
+register_type(float, widget_type="FloatSlider")
+register_type(int, widget_type="Slider")
+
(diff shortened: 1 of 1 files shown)


tlambert03:
this is just a lovely example.  thank you @glyg!

Two thoughts:
1. if you add these lines at the top, you can turn everything into sliders (which are more fun than spinners? what do you think?)
```py
from magicgui import magicgui, widgets, register_type

register_type(float, widget_type='FloatSlider')
register_type(int, widget_type='Slider')
```
Untitled

2. the default range of 0-1000 isn't always appropriate for each parameter, but at the same time it's really nice that the signal functions are not decorated with magicgui here.  So, just as an FYI, if you want, you can embed widget options in a parameter's type annotation with `Annotated` like this:

```py
from typing_extensions import Annotated

Freq = Annotated[float, {"min": 0, "max": 2 * np.pi}]
# repeat as desired for other common param types

def sine(
    duration: float = 10.0, size: int = 500, freq: Freq = 0.5, phase: float = 0.0
) -> Signal:
    ...

```

glyg:
Sliders are indeed more fun!
The `Annotated` type is a great trick, I was thinking to define an options dictionary for each signal, but that's way better.

I'll update the example soon-ish

tlambert03 (approved):
thanks @glyg!  looks awesome

glyg:
\o/  Thanks a lot!

PR #331, opened 2021-12-09
Relay signals from Container to Container

hanjinliu:
Looks like we only have to change this single line.
I think it's reasonable in most cases to send signal to the ancestors.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -11,7 +11,6 @@ on:
       - main
   workflow_dispatch:
 
-
 jobs:
   test:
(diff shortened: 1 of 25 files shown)


tlambert03:
yep looks great!  Could you add a small test?  perhaps create a `SliceEdit` widget, connect it's changed signal to a `unittest.mock.Mock` object, then directly change one of the subwidgets in the SliceEdit widget and assert that the mock has been called?  (should be patterns of this elsewhere in the tests, let me know if you can't find any examples).

thanks!

hanjinliu:
Like this? It tests that the parent `Container` can observe the value change event of a `SliceEdit`.

tlambert03 (approved):
thanks @hanjinliu!

PR #342, opened 2022-01-01
Support sequences

hanjinliu:
Trying to add typing supports of such as `list[str]` and `tuple[int, int]`.
These examples are how it is intended to be used.

```python
@magicgui
def random_2d_image(shape: tuple[int, int] = (100, 100), max: float = 1) -> ImageData:
    return np.random.random(shape) * max
```

```python
@magicgui
def gaussian_filter(img: ImageData, sigma: list[float] = [1.,]) -> ImageData:
    if len(sigma) == 1:
        sigma = sigma[0]
    return skimage.filters.gaussian(img, sigma)
```

Still need more strict typing to pass the tests.


tlambert03:
i like it! thanks!  let me know when you're ready for a full review

hanjinliu:
Hi @tlambert03,
I think things are almost done. Could you check my PR?

A few things that I cannot decide by myself are:
- Should we also support typing like `Sequence[int]` or `Iterable[int]`? The new type matcher only detects `list[...]` and `tuple[...]` (and their subclasses).
- For `ListEdit`, widget options of the child widgets can be set using "options" argument. For instance, options of the `int` in `list[int]` is editable by `ListEdit(value=[0, 1, 2], options={"max": 10})`. But there is no elegant way for `TupleEdit` since its child widgets could be different (like `tuple[int, str]`).

hanjinliu:
Sorry, @tlambert03, the last commit is a separated one... (concerning #361)
Can I split them into another PR in some way?

tlambert03:
> Sorry, @tlambert03, the last commit is a separated one... (concerning https://github.com/napari/magicgui/issues/361)
Can I split them into another PR in some way?

Hey @hanjinliu, sorry for the delay here.  I'm ready to take a look at this.  

Note that https://github.com/napari/magicgui/pull/362 is solving #361, and a lot of other issues with forward refs.  But will likely create conflicts here.  Perhaps we should merge that one before coming back to the type_map updates in this PR?  I can help with that merge if you'd like.

I'll do a more thorough review once that is resolved, and we can get this in for the next release, which I'd like to do soon.

hanjinliu:
Hi @tlambert03, I've read the upcoming changes in type mapping.
I have no idea how I can pass the "segmentation fault" thing... could you help me solve this?

hanjinliu:
I was misreading a little bit few days ago... but I updated things for the new `TypeWrapper`.
I hope I'm not doing anything wrong.

gselzer:
@hanjinliu this is super awesome! I'd love to use this in my plugin, what is the status of this PR?

tlambert03:
> what is the status of this PR?

waiting on me/review at this point

tlambert03:
so sorry for the delay on this @hanjinliu.  been busy and haven't found the time to give this the close look I wanted to.  Since it's pretty much just a new feature, (and at first glance looks to be nicely implemented!)  I'm going to just merge this so people like @gselzer can take a crack at it.  We can iterate as needed if anything arises.  Thanks again for your contribution and your patience!

hanjinliu:
Hi, @tlambert03 
Thank you for the full review! Yes, I totally agree with that. Let's see how it goes well with plugin development.

gselzer:
@hanjinliu @tlambert03 this seems to work great! Thanks for working on this!

This allows us to finally set up variable length arrays and lists in napari-imagej!

Here's an early look at a N-Dimensional gaussian blur with separate dimensional sigmas:

image

PR #355, opened 2022-01-18
Create return matcher for extensible return widget creation

gselzer:
This PR introduces a class `ReturnMatcher`, an analogue of `TypeMatcher` for widgets displaying function outputs.

As of now, there are only two `ReturnMatcher`s written. One supports "simple" types via `widgets.LineEdit`. The other supports "tabular" types via `widgets.Table`. There is nothing preventing us from writing more of these to support other types.

The default return widget is a `widgets.LineEdit`, chosen to best align with previous behavior.


tlambert03:
generally looks good!  I think, if possible, I'd prefer not to have an entirely different module for return annotations.  If you can point to the specific places that made you feel like you needed a fully different `get_widget_class` function and new module, I might be able to suggest solutions that would prevent all the code duplication

gselzer:
> If you can point to the specific places that made you feel like you needed a fully different `get_widget_class` function and new module, I might be able to suggest solutions that would prevent all the code duplication

I don't know that there were any, that might just be my java background showing 😅

If we propagate the `is_input` parameter into `type_map.py`, we should be able to merge the two classes. I'll do that quick.

Do you want to leave the `TypeMatcher` class the way it is, or should we call it `ParamMatcher` to align with the `ReturnMatcher` class?

tlambert03:
> If we propagate the is_input parameter into type_map.py, we should be able to merge the two classes. I'll do that quick.

even more, can you think of specific scenarios where we want to map a given type to a different widget on the output than we would on the input?  Most of the type->widget mappings kind of make sense for the value type (independent of whether it's an input or an output).  The one exception might be that outputs should be "enabled=True"

gselzer:
> can you think of specific scenarios where we want to map a given type to a different widget on the output than we would on the input?

Sure, I mean one I think makes the least sense would be a `bool`, where you'd want a check box if input, but that probably doesn't make a ton of sense for the output.

Similarly, you might want a slider widget for a numerical input, but I don't *think* you'd want that for the output

tlambert03:
> Sure, I mean one I think makes the least sense would be a bool, where you'd want a check box if input, but that probably doesn't make a ton of sense for the output.

doesn't it though?  If someone wrote:

```
@magicgui(result_widget=True)
def some_widget(...) -> bool:
    ...
```

wouldn't a checkbox that ticks and unticks itself as the result changed from True to false be as reasonable of a default as any?  I see that you're using `LineEdit` which is also fine... but I think we need to have a relatively high bar here to justify a totally different type matcher.

I do think that anything that requires further input (like the filedialog for pathlib.Path) _is_ a good example of something that doesn't make sense as an output.

Lemme ponder this one a bit :)

gselzer:
@tlambert03 I condensed the two type matchers into one module, but there are still two matchers. I did actually find a difference between the two that wasn't just a difference in return widget; I get test failures in other tests if I use [this block](https://github.com/gselzer/magicgui/blob/9ec44a37bdd0c246d1bd93eba9b1434f61d1190e/magicgui/type_map.py#L229) in return widget matching.

magicgui/types.py
```diff
@@ -24,6 +24,9 @@
 #: A function that takes a ``(value, annotation)`` argument and returns an optional
 #: :attr:`WidgetTuple`
 TypeMatcher = Callable[[Any, Optional[Type]], Optional[WidgetTuple]]
+#: A function that takes a ``(value, annotation)`` argument and returns an optional
+#: :attr:`WidgetTuple`
+ReturnMatcher = Callable[[Any, Optional[Type]], Optional[WidgetTuple]]
```
tlambert03, line 29: since this is just an alias (not an actual class or anything that could be used for `isinstance` for example) and since this alias is exactly the same as `TypeMatcher`.  Let's just use `TypeMatcher` for both.

magicgui/type_map.py
```diff
@@ -43,6 +44,7 @@ class MissingWidget(RuntimeError):
 
 
 _RETURN_CALLBACKS: DefaultDict[type, list[ReturnCallback]] = defaultdict(list)
+_RETURN_MATCHERS: list[ReturnMatcher] = list()
```
tlambert03, line 47: see comment in `types.py` ... let's just use `_RETURN_MATCHERS: list[TypeMatcher] = list()`

magicgui/type_map.py
```diff
@@ -112,6 +114,19 @@ def type_matcher(func: TypeMatcher) -> TypeMatcher:
     return func
 
 
+def return_matcher(func: ReturnMatcher) -> ReturnMatcher:
```
tlambert03, line 117:
```suggestion
def return_matcher(func: TypeMatcher) -> TypeMatcher:
```

magicgui/type_map.py
```diff
@@ -112,6 +114,19 @@ def type_matcher(func: TypeMatcher) -> TypeMatcher:
...
+    """Add function to the set of return matchers.
+
+    Example
+    -------
+    >>> @return
```
tlambert03, line 122:
```suggestion
    >>> @return_matcher
```

magicgui/type_map.py
```diff
@@ -225,9 +240,90 @@ def pick_widget_type(
...
+            return _cls, {**options, **opts}  # type: ignore
+
+    # Chosen for backwards/test compatibility
+    return widgets.LineEdit, {"gui_only": True}
+
+
 def get_widget_class(
     value: Any = None,
     annotation: type | None = None,
+    is_input: bool = None,
```
tlambert03, line 326: this is technically a breaking API change, since this is a public function and this changes the order of positional args.  (not to mention the fact that the current default of `is_input=None` will change this function to return widgets by default). Can we put this after `options`, change the name to `is_result` (to match the current wording used elsewhere of `result_widget`, and use a default value of `is_result=False`?

magicgui/widgets/_bases/create_widget.py
```diff
@@ -20,6 +20,7 @@ def create_widget(
     label=None,
     gui_only=False,
     app=None,
+    is_input: bool = True,
```
tlambert03, line 23: see other comment, move after `options` and change to `is_result=False`

tlambert03 (commented):
this is looking good @gselzer!    Thank you also for the lovely tests.

I've made few minor comments about naming and API.  But other than that I think it's very close!

tlambert03:
also, looks like there's one last failing test: https://github.com/napari/magicgui/runs/4913849327?check_suite_focus=true#step:6:296

gselzer:
> also, looks like there's one last failing test: https://github.com/napari/magicgui/runs/4913849327?check_suite_focus=true#step:6:296

@tlambert03 I don't understand that test failure. Why is a function in `test_magicgui.py` even being called?

magicgui/type_map.py
gselzer: Change implemented in 603ebd7

magicgui/type_map.py
gselzer: Change implemented in 603ebd7

magicgui/type_map.py
gselzer: Change implemented in 603ebd7

magicgui/type_map.py
gselzer: I made the suggested change and rebased it into the commit that added this so that we wouldn't have any commits where `is_input` existed. Please let me know what you think.

magicgui/widgets/_bases/create_widget.py
gselzer: I made the suggested change and rebased it into the commit that added this so that we wouldn't have any commits where `is_input` existed. Please let me know what you think.

tlambert03:
> > also, looks like there's one last failing test: https://github.com/napari/magicgui/runs/4913849327?check_suite_focus=true#step:6:296
> 
> @tlambert03 I don't understand that test failure. Why is a function in `test_magicgui.py` even being called?

ah, because of a poorly designed test causing side effects over there :)

in line ~446 of test_magicgui.py, add this try/finally, and pop the registered types.

```py
    register_type(int, return_callback=check_value)
    register_type(Base, return_callback=check_value)

    try:
        @magicgui
        def func(a=1) -> int:
            return a

        func()
        with pytest.raises(AssertionError):
            func(3)

        @magicgui
        def func2(a=1) -> Sub:
            return a

        func2()
    finally:
        from magicgui.type_map import _RETURN_CALLBACKS
        _RETURN_CALLBACKS.pop(int)
        _RETURN_CALLBACKS.pop(Base)
```

(if this happens again, I can deal with this on a more global level)

gselzer:
Gotcha, that seems to have fixed it on my side @tlambert03. Let me know what you think

tlambert03 (approved):
Thanks a lot for this @gselzer!  very handy addition!

PR #362, opened 2022-01-29
Add new type normalization TypeWrapper, fix builtin and optional ForwardRefs

### LINKED ISSUE #361: Cannot resolve ForwardRef if `from __future__ import annotation` is used.
In a .py file, the type matcher fails to convert simple types (`int`, `str`, ...) into widgets if the file starts with `from __future__ import annotation`.

```python
from __future__ import annotation
from magicgui import magicgui

@magicgui
def f(i: int):
    pass
```

Looks like it is because `int` is converted into `"int"`. Additional matching system is needed.

tlambert03:
Keeping forward references working well has been challenging, partly because it's hard, and partly because the type normalization logic was a bit scattered – requiring you to call the same "normalize_type" function in multiple places – and partly because typing is just hard – with pretty minimal help from the typing module itself in terms of runtime checking. 

This PR borrows quite a bit of logic from [pydantic's ModelField](https://github.com/samuelcolvin/pydantic/blob/9d631a3429a66f30742c1a52c94ac18ec6ba848d/pydantic/fields.py#L564-L565) (which has had to deal with similar problems).  Specifically, it adds a `TypeWrapper` class that does the job of receiving a type in any form, recursing through builtin generics like Unions, Optional, and containers, and standardizing it all into a common interface.

```py
In [1]: from magicgui._type_wrapper import TypeWrapper

In [2]: from typing import Optional

In [3]: tw = TypeWrapper(Optional["napari.layers.Layer"])

In [4]: tw.is_resolved
Out[4]: False

In [5]: tw.resolve()
Out[5]: napari.layers.base.base.Layer

In [6]: tw.is_resolved
Out[6]: True

In [7]: tw
Out[7]: TypeWrapper(type='Optional[Layer]', required=False, default=None)

In [8]: tw.outer_type_
Out[8]: napari.layers.base.base.Layer
```

This gets rid of a lot of ugly logic in `type_map.py`.  And, because the forward ref resolution recurses through generics, it should prevent issues like [#affinder](https://github.com/jni/affinder/pull/50) (where `Optional['forward.ref']` didn't work).

This also closes #361 

Other things happening in this PR:

- adding a test to pull in napari's `test_magicgui.py` file on every PR
- removing `dict`, `list`, and `tuple` from return annotations that create Table result widgets  (cc @gselzer... I think those are bit too open-ended, because most dicts and lists will not be valid for table input.  perhaps someday we can have `Dict[str, List[Any]]` work for a table, but `dict` in general is insufficient.  `np.ndarray` and `pd.DataFrame` will still work, as will `magicgui.widgets.Table`.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -54,10 +54,11 @@ jobs:
       - name: Test with tox
         uses: GabrielBB/xvfb-action@v1
         with:
-          run: python -m tox
+          run: tox
         env:
           PLATFORM: ${{ matrix.platform }}
           BACKEND: ${{ matrix.backend }}
+
       - name: Coverage
         if: runner.os == 'Linux'
         uses: codecov/codecov-action@v2
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -48,11 +48,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools tox tox-gh-actions
+          pip install -e .[testing]
+          pip install ${{ matrix.backend }}
+
       - name: Test with tox
         uses: GabrielBB/xvfb-action@v1
         with:
-          run: tox
+          run: pytest -v --color=yes --cov-report=xml --cov=magicgui --basetemp={envtmpdir} {posargs}
         env:
           PLATFORM: ${{ matrix.platform }}
           BACKEND: ${{ matrix.backend }}
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -48,13 +48,12 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[testing]
-          pip install ${{ matrix.backend }}
+          pip install setuptools tox tox-gh-actions
 
       - name: Test with tox
         uses: GabrielBB/xvfb-action@v1
         with:
-          run: tox
+          run: python -m tox
         env:
           PLATFORM: ${{ matrix.platform }}
           BACKEND: ${{ matrix.backend }}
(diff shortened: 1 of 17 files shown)


jni:
"details later" :rofl:

tlambert03:
@jni, "details" have been added ;)

gselzer:
> * removing `dict`, `list`, and `tuple` from return annotations that create Table result widgets  (cc @gselzer... I think those are bit too open-ended, because most dicts and lists will not be valid for table input.  perhaps someday we can have `Dict[str, List[Any]]` work for a table, but `dict` in general is insufficient.  `np.ndarray` and `pd.DataFrame` will still work, as will `magicgui.widgets.Table`.

@tlambert03, yeah, I agree that generally speaking they are too broad. Do you have any plans for the widget type you want them to map to?

tlambert03:
>  Do you have any plans for the widget type you want them to map to?

not really, but naively, if someone simple annotated something as returning a dict, I think "mapping", and that's probably just a form layout with key on left, and stringified version of value on right

gselzer:
Sure, but is that not just a table? Maybe it is worth writing a `Table` subclass that specializes in displaying `dict`s, `list`s, `tuple`s that do not conform to the dimensionality needed for a table?

I agree with the change regardless, I just am wondering how we support these things with minimal additional effort :grin:

tlambert03:
> Sure, but is that not just a table?

visually, yes... but you can't literally just pass their dict and expect it to work, so no...  
I saw that you previously had a test that "tests that it fails" if the dict is incorrectly formatted, but I think failing only after someone creates a widget and tries to populate it is not the correct approach.  Basically, we just shouldn't offer this at all until we've actually implemented the conversion

gselzer:
Ah, interesting, I didn't know about `Container`.  Good to know.

Does `Container` have any API for returning that data, or would you have to parse through each of the `LineEdit`s if you wanted to get the `dict` or other structure back from the `Container`?

tlambert03:
> Does `Container` have any API for returning that data, or would you have to parse through each of the `LineEdit`s if you wanted to get the `dict` or other structure back from the `Container`?

```py
{w.label:w.value for w in container}
```

gselzer:
Makes sense. Thanks for sating my curiosity!

PR #368, opened 2022-02-11
Remove event deprecations strategy (for release 0.4.0)

tlambert03:
This PR removes the `event.py` module that was mediating the deprecation strategy introduced in #253.  This will make magicgui less sensitive to changes in private attributes over in psygnal.  The deprecation strategy has been in place for 4 months.

This is a breaking change, so the next release will be 0.4.0.


aeisenbarth:
These days I noticed that the events.py broke psygnal's `connect_setattr` when used in MagicGUIs. The attribute setter receives an Event instead of the new attribute value. This is because `_new_callback` does not contain True for the setattr callback. I was just about to create a patch and then discovered this recent PR.

I then ran my test case with this PR and it also fixes `connect_setattr`!
Just mentioning that I'm looking forward to the release.

tlambert03:
Thanks for taking the time to let me know @aeisenbarth!  I think we're pretty much in the "safe" zone for merging and releasing this with 0.4.0... likely in a couple weeks or so.

uschmidt83:
I just noticed that magicgui 0.4.0 broke `stardist-napari`. How could I have avoided that? I've never seen a deprecation warning related to this.

uschmidt83:
Wait a second, you opened this PR https://github.com/stardist/stardist-napari/pull/14 to fix this in our plugin.
We just haven't made a new release since then...

uschmidt83:
Ok, I'm very confused now... stardist-napari 2022.4.8 already includes your PR to remove magicgui.events. However, stardist-napari is downgraded to 2022.4.4 in our CI tests: https://github.com/stardist/stardist-napari/actions/runs/2386252414

Sorry about the rambling, feel free to delete all of this.

PR #379, opened 2022-02-22
Model-based widget creation

tlambert03:
quite similar to #318 ... this pattern would use pydantic directly in widget declaration:
```python
from magicgui import Field, GUIModel
from magicgui.widgets import Slider

# this is a pydantic subclass
class MyWidget(GUIModel):
    x: int = 1
    # magicgui kwargs are prefaced by ui_
    y: int = Field(2, ui_widget_type=Slider)

    class Config:
        #container kwargs in the config
        layout = "horizontal"


obj = MyWidget()
obj.x  # values are directly on the object, like pydantic
obj.events.x.connect  # events are in a psygnal group
obj.gui.show(run=True)  # the widget is at the `.gui` attribute
```


tlambert03:
closing this in favor of #446

PR #386, opened 2022-03-18
Fix pyqt conda isntall instruction

dstansby:
On `conda-forge` the package is `pyqt`, not `pyqt5`. Perhaps this should be changed to `pyqt=5` instead?


tlambert03 (approved):
indeed!  thank you!

Issue #381, opened 2022-03-03
Add ability to block call button until call has finished


dstansby:
It's currently possible to press the call button several times in a row, which will launch multiple runs one after the other - see my toy example below. For computationally intensive tasks it would be nice to be able to 'block' any more pressing of the call button if the main function is currently being run.

```python
from magicgui import magicgui
import time

@magicgui(call_button="Run")
def run_timer():
    print('Asleep')
    time.sleep(2)
    print('Awake')

run_timer.show()
```

dstansby:
It looks like this is already enabled:
https://github.com/napari/magicgui/blob/0fa849adba0c1e15eb850f042e5fc5c7cceff362/magicgui/widgets/_function_gui.py#L194-L202

But in my original use case the function I was using was using a [napari thread worker](https://napari.org/guides/threading.html#threading-in-napari-with-thread-worker), so I presume on pressing the button the function was running but not blocking execution, so the run button was immediately being re-enabled. It should be possible to add some special case handling to the above lines of code to block and wait for the thread worker to complete.

tlambert03:
Thanks, I thought I had added something like that 😂

I'm actually not sure about the special casing... there are many ways that someone could spin up a thread inside of a function call (beyond thread_worker), so magic gui would need to do some pretty gnarly introspection.  By contrast, the end user could implement this quite easily with a similar button blocker than gets unlifted on thread finish.  (Remember that magicgui has no knowledge of napari, it's the other way around).

I can make an example of that if it would help.

dstansby:
👍 - to double check, the recommendation is to not use the "call_button" argument, and create a custom button with custom blocking behvaviour? If so that sounds good to me, and I guess this issue can be closed?

tlambert03:
Sorry no, I meant you can use the same call button (we should make it public), but set it to disabled, and then connect to the threadworker finished signal to trigger a callback that re-enables the button. 

Let me try to create an example later today.

tlambert03:
since this issue has already been solved for synchronous code, and an async solution is out of scope for the near-term (and there is a workaround), I'm going to close this.  but feel free to open if new issues arise

PR #391, opened 2022-03-29
Make call_button public

dstansby:
See https://github.com/napari/magicgui/issues/381#issuecomment-1060550957 and associated discussion. This allows public access to the button automatically created by `magicgui` so it can be modified after creating the GUI.

### DIFF
--- a/magicgui/widgets/_function_gui.py
+++ b/magicgui/widgets/_function_gui.py
@@ -79,6 +79,7 @@ class FunctionGui(Container, Generic[_R]):
         If True, create an additional button that calls the original function when
         clicked.  If a ``str``, set the button text. by default False when
         auto_call is True, and True otherwise.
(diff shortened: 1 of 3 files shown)


tlambert03:
thanks @dstansby.  just some outdated tests left that need fixing.

magicgui/widgets/_function_gui.py
```diff
@@ -186,23 +187,23 @@ def __init__(
 
         if call_button is None:
             call_button = not auto_call
-        self._call_button: PushButton | None = None
+        self.call_button: PushButton | None = None
```
tlambert03, line 190: actually, i might prefer a public accessor rather than changing the private attr.

```python
@property
def call_button(self) -> Optional[PushButton]:
    return self._call_button
```

Issue #372, opened 2022-02-15
All value-like Container widgets should inherit from ValueWidget


tlambert03:
see https://forum.image.sc/t/using-layer-properties-to-pass-variables/63342/6

PR #396, opened 2022-04-06
Fix bug in creation of RangeEdit using create_widget

Czaki:
When run this code 

```python
from magicgui.widgets import create_widget

w = create_widget(annotation=range)
```
it ends with error:
```
  File "/home/czaki/Projekty/neutrofile/Tutorial/sample_magicgui.py", line 3, in <module>
    w = create_widget(annotation=range)
  File "/home/czaki/.pyenv/versions/3.8.3/envs/neutrofile/lib/python3.8/site-packages/magicgui/widgets/_bases/create_widget.py", line 96, in create_widget
    widget = wdg_class(**kwargs)
  File "/home/czaki/.pyenv/versions/3.8.3/envs/neutrofile/lib/python3.8/site-packages/magicgui/widgets/_concrete.py", line 558, in __init__
    raise TypeError(f"Invalid value type for {type(self)}: {type(value)}")
TypeError: Invalid value type for <class 'magicgui.widgets.RangeEdit'>: <class 'magicgui.widgets._bases.value_widget._Unset'>
```

When fixing this error by checking if `value` is `None` or `UNSET` then also happens problem with an unexpected `nullable` field.
The second error I fix  by `pop` the `nullable` from kwargs. But I'm not sure if it is the correct solution.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
       - id: pyupgrade
         args: ["--py37-plus", "--keep-runtime-typing"]
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
   - repo: https://github.com/PyCQA/flake8
--- a/magicgui/widgets/_concrete.py
+++ b/magicgui/widgets/_concrete.py
@@ -553,7 +553,7 @@ class RangeEdit(Container):
         **kwargs,
     ):
         value = kwargs.pop("value", None)
-        if value is not None:
+        if value is not None and value is not UNSET:
             if not all(hasattr(value, x) for x in ("start", "stop", "step")):
                 raise TypeError(f"Invalid value type for {type(self)}: {type(value)}")
             start, stop, step = value.start, value.stop, value.step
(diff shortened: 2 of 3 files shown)


tlambert03:
thanks @Czaki!  These changes make sense / work for me.  more broadly, I think the issue here stems from having these "container-based" classes that more or less implement the `ValueWidget` protocol, but don't directly inherit from valuewidget.  (see https://github.com/napari/magicgui/issues/372).

This change is totally fine with me as a first step... maybe in the future, I can get around to making them actually be `ValueWidgets` (in which case `nullable` should work again, and allow things like `create_widget(annotation=Optional[range])`

tlambert03 (approved):

tlambert03:
This came up before in some issue, can't find it atm, since not all widgets are necessarily value widgets, I'm not sure that's the proper annotation?

Czaki:
Hm. In general I know because Container is also widget. But what input to `create_widget` could create non-value widget?

Or you think about user registered widget?

tlambert03:
> But what input to create_widget could create non-value widget?

Yeah, _thats_ a good question :)

I can't immediately think of one.  I'm open to it.  I remember trying to change it one point and then realizing that changing that annotation requires a lot of other type changes elsewhere, but it might be worth it.  Feel free to dig a bit deeper if you're interested

Czaki:
If I should fix somehow failing test? It looks non-related.

PR #397, opened 2022-04-07
Fix bug with small default range in SpinBox

Czaki:
The default range in SpinBox is 0, 1000 which is really small. Try to create a widget with a value outside the default range causing the Exception. This PR adjusts the min-max range if the initial value is outside it.

### DIFF
--- a/magicgui/widgets/_bases/ranged_widget.py
+++ b/magicgui/widgets/_bases/ranged_widget.py
@@ -45,7 +45,7 @@ class RangedWidget(ValueWidget):
         val = kwargs.pop("value", UNSET)
         super().__init__(**kwargs)
 
-        tmp_val = val if val not in (UNSET, None) else 1
+        tmp_val = float(val if val not in (UNSET, None) else 1)
 
         self.step = step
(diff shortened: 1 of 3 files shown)


Czaki:
I'm not sure how to fix this error. The simplest method will be to hide arguments in kwargs and maybe overload?

Czaki:
Thanks :)

tlambert03 (approved):
thank you @Czaki!

magicgui/widgets/_bases/ranged_widget.py
```diff
@@ -1,28 +1,37 @@
...
 
     Parameters
     ----------
     min : float, optional
-        The minimum allowable value, by default 0
+        The minimum allowable value, by default 0 (or `value` if `value` is less than 0)
     max : float, optional
-        The maximum allowable value, by default 1000
+        The maximum allowable value, by default 1000 (or `value` if `value` is greater
+        than 1000)
```
Czaki, line 21: ahh. I do not update this line. The default value is 999

PR #403, opened 2022-05-04
Change to use adaptive step is SpinBox and FloatSpinBox

Czaki:
To provide better user experience set to use adaptative step by default is SpinBox and FloatSpinBox


tlambert03 (approved):
thanks!

PR #404, opened 2022-05-09
Bug fix of type check in ListEdit

hanjinliu:
This PR fixes a bug in `ListEdit`.
I found `ListEdit` could not deal with type annotation of `list[...]` with an empty list as the default value.

```python
@magicgui
def f(x: list[int]): ...  # OK

@magicgui
def f(x: list[int] = []): ...  # Error!

```


tlambert03 (approved):
thanks!

Issue #406, opened 2022-05-10
Resetting choices triggers too many changed events


aeisenbarth:
**Describe the bug**
When resetting choices on a ComboBox that had hundreds of items, the current choices are first sequentially removed, then the new choices are set e.g. from a callable. If the currently selected choice was the last item, every removal will change the selection and trigger a change event on the widget.

**To Reproduce**
1. Create a combobox with choices from a callable.
2. Set the value to any but the first choice.
3. Add listen to the `changed` event.
4. Reset choices.
5. Count how many times the event listener was called.

**Expected behavior**
I would think event listeners are only interested in the resulting state after resetting choices. Resetting choices should be an atomic change and trigger a single event. 

**Environment (please complete the following information):**
 - OS: Ubuntu 21.10
 - backend: PyQt5==5.15.6
 - magicgui version: >=0.3.7

PR #407, opened 2022-05-10
Block event emits during choices batch update

aeisenbarth:
This PR fixes https://github.com/napari/magicgui/issues/406.

Every `_qwidget.removeItem(i)` can potentially change the `currentIndex` and trigger an event.
By blocking events during the remove/add operations, we turn this into a batch change and trigger a single event with the final `currentIndex` afterwards.

### DIFF
--- a/magicgui/backends/_qtpy/widgets.py
+++ b/magicgui/backends/_qtpy/widgets.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import math
 import re
+from contextlib import contextmanager
 from functools import partial
 from typing import TYPE_CHECKING, Any, Iterable, Sequence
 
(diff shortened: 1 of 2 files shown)


tlambert03 (commented):
thank you @aeisenbarth! 

can I make a request?  I know we've used this `blockSignals(True) blockSignals(False)` pattern elsewhere in the file, but it should really be in a context manager (in case an exception occurs in between... otherwise it would be left in a blocked state).

Let's add a helper function at the top:

```python
from contextlib import contextmanager

@contextmanager
def _signals_blocked(obj: QtW.QWidget):
    before = obj.blockSignals(True)
    try:
        yield
    finally:
        obj.blockSignals(before)
```

and then use it as:

```python
with _signals_blocked(self._qwidget):
    ...
```

aeisenbarth:
Hi Talley,
I knew the psygnal context manager, but this here seemed to be a different signaling mechanism.
I added the context manager for Qt, and also checked for similar occurrences where multiple signals were emitted.

tlambert03 (approved):
thanks!

PR #415, opened 2022-06-08
Return empty tuple for no selected files

aeisenbarth:
This change makes the FileEdit in EXISTING_FILES mode return an empty tuple if no files are selected.

The ability to detect "no selected file" is important to prevent parsing invalid/non existing files. The FileEdit has three possibilities, either it always has a Path value, or it is nullable and the neutral value should be None, or it is in multiple file selection mode and according to the signature the neutral value should be a tuple of zero length.

However, since `"".split(some_delimiter)` always gives `[""]` instead of `[]`, it currently returns a tuple with an empty path `Path(".")` which equals the home directory and is rather impractical to distinguish from an intentionally selected path.

### DIFF
--- a/magicgui/widgets/_concrete.py
+++ b/magicgui/widgets/_concrete.py
@@ -510,13 +510,15 @@ class FileEdit(Container):
         """Set current file path."""
         if value is None and self._nullable:
             value = ""
-        if isinstance(value, (list, tuple)):
-            value = ", ".join(os.fspath(p) for p in value)
-        if not isinstance(value, (str, Path)):
+        elif isinstance(value, (list, tuple)):
+            value = ", ".join(os.fspath(Path(p).expanduser().absolute()) for p in value)
+        elif isinstance(value, (str, Path)):
+            value = os.fspath(Path(value).expanduser().absolute())
+        else:
             raise TypeError(
                 f"value must be a string, or list/tuple of strings, got {type(value)}"
             )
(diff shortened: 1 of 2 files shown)


tlambert03:
thanks!  can you add a test somewhere?  Also, what about `tuple(Path(p) for p in text.split(", ") if p.strip())` ... which should additionally catch any `""` that happen to sneak through with other valid paths?

aeisenbarth:
Good point! I was looking at the code that explicitly sets the line edit to an empty string. My version wouldn't have caught spaces `" "`. I don't see yet how that could happen other than by user input (but then there would be more cases to consider, like files separated by extra spaces `",  "` and when stripping those we can't support file paths ending with space and I believe file paths containing comma and space would already mess it up).

aeisenbarth:
I also fixed two more issues:
- For a single (nullable) path, `""` would be expanded to an absolute path, the same for multiple mode with an empty files list.
- For multiple files, only the first path was expanded, since the expansion was applied after joining to a single string.

Also, is `str` still allowed for assignmend? The mypy hook complained because the annotation is `Path` (although the TypeError in the code allows `str`). Should I change annotations to `PathLike`?

tlambert03 (commented):
thanks! this looks great!  and thank you for the tests!

> The mypy hook complained because the annotation is Path (although the TypeError in the code allows str). Should I change annotations to PathLike?

which annotation?  in the getter?  (things look good to me as is)

PR #416, opened 2022-06-10
feat: add show_values_dialog [WIP]

tlambert03:
adds `magicgui.widgets.show_values_dialog` which lets you show a modal dialog to retrieve a set of values from the user 

```python
In [1]: from magicgui.widgets import show_values_dialog

In [2]: show_values_dialog(param_a=int, param_b=str, title='asdfsd')
Out[2]: {'param_a': 0, 'param_b': ''}
```

Screen Shot 2022-06-09 at 8 13 51 PM


tlambert03:
@ctrueden and @gselzer, this is partially in response to some recent discussions.  Lemme know what you think of the API

gselzer:
@tlambert03 this is cool, and the API is pretty succinct! The documentation for the widget type is nice, too.

Do you have some use cases in mind?

tlambert03:
> Do you have some use cases in mind?

yes, i got thinking about this from the `New Image` discussion.  For example, if you needed to collect some input (size, content, dtype, etc) to create a new image.

In general, this could be used by napari anytime a command requires parameters

gselzer:
Gotcha, thanks.

I think I'm just having a bit of trouble thinking through when to use `request_values` over a plain call to `magicgui`. I think this `request_values` thing is pretty lightweight in that it can't do anything with the values, right?

Since the major benefit of this functionality is convenience, I keep wondering if it would be useful to allow a postprocessing routine in the signature.

For example, if you picked up the size, content, dtype, etc. for the new image command, you could pass a function to `request_values` that, after value selection, turned all of those values into an image, and that's the only output you get from the `request_values` call. Or in the case of napari-imagej you could have a postprocessing function that turns all of the python values into their java equivalents.

But that seems to close to what `magicgui` is supposed to do? Maybe we need a section on when to use this vs. `magicgui`.

tlambert03:
> I think I'm just having a bit of trouble thinking through when to use request_values over a plain call to magicgui. I think this request_values thing is pretty lightweight in that it can't do anything with the values, right?

yep, much more lightweight.  makes no assumptions about what someone is doing with the values downstream.  In general actually, I kinda think `@magicgui` itself is being abused/overused a bit, as the "only" way to retrieve values from a user.  So I'd like to gradually build some lower level stuff that gets used more often.

> I keep wondering if it would be useful to allow a postprocessing routine in the signature.

I can see the potential usefulness there.  maybe for a follow up PR though.  Currently, since request_values blocks, you would just achieve it as follows:

```
do_something_with_values(**request_values(...))
```

gselzer:
> In general actually, I kinda think `@magicgui` itself is being abused/overused a bit, as the "only" way to retrieve values from a user.

I definitely agree, and may be guilty of this in napari-imagej :sweat_smile: 

b

> I can see the potential usefulness there. maybe for a follow up PR though. Currently, since request_values blocks, you would just achieve it as follows:
> 
> ```
> do_something_with_values(**request_values(...))
> ```

Yup, definitely should not go in this PR, especially as I cannot imagine how I'd want the API for that to look. Do you need to specify the output types of the `Callable` in the signature somehow? I'd think you should, especially if the `Callable` alters the number of inputs you are returning, but how would that look best?

Maybe we could add a modal option for `magicgui`...

PR #422, opened 2022-06-14
Emit signal only if value of caterogical widget changed

Czaki:
This is bugfix PR for not emmiting signal that value is changed if nothing changed in selection.


tlambert03 (approved):
This looks good to me.  Sorry for the relatively convoluted stuff going on here.  I can merge this and cut 0.5.1 if you're happy with it

PR #425, opened 2022-06-16
Turn off adaptive step if "step" option is given

hanjinliu:
This PR changes the default behavior of `step` and `adaptive_step`.

I like the idea of using adaptive step by default but it should be turned off if `step` is explicitly given (also in terms of backward compatibility).
It is not straightforward that the `step` option is ignored unless `"adaptive_step": False` is specified at the same time.

```python
@magicgui(x={"step": 0.1})  # ignored
def func(x: float): ...

@magicgui(x={"step": 0.1, "adaptive_step": False})  # OK
def func(x: float): ...
```

To do this, I corresponded `step=None` to adaptive step mode, and accordingly, the default value of `step` is changed from `1` to `UNSET` and `step` returns `None` if the widget is in adaptive step mode. 
This change might violate compatibility if one checks the value like `widget.step == 1` but it seems very rare.

### DIFF
--- a/magicgui/widgets/_bases/ranged_widget.py
+++ b/magicgui/widgets/_bases/ranged_widget.py
@@ -119,7 +119,7 @@ class RangedWidget(ValueWidget):
     @property
     def adaptive_step(self):
         """Whether the step size is adaptive."""
-        return self._widget._mgui_get_adaptive_step()
+        return self.step is None
 
     @adaptive_step.setter
     def adaptive_step(self, value: bool):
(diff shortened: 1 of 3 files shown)


magicgui/widgets/_bases/ranged_widget.py
```diff
@@ -29,8 +29,7 @@ def __init__(
         self,
         min: Union[float, _Unset] = UNSET,
         max: Union[float, _Unset] = UNSET,
-        step: float = 1,
-        adaptive_step: bool = True,
+        step: Union[float, _Unset, None] = UNSET,
```
Czaki, line 32: Could you motivate me to why you add `None` and `_Unset` here? One of them should be enough.

Czaki (approved):

magicgui/widgets/_bases/ranged_widget.py
hanjinliu: That's because min and max are also initialized with UNSET... I guess UNSET is for `value` only (because it might be set to None intentionally) so should we just change the defaults of min, max and step to None?

PR #429, opened 2022-06-20
Expose QScrollArea as native widget

dstansby:
Fixes https://github.com/napari/magicgui/issues/431 - this allows an external user to access the `QScrollArea` by returning it as the native widget. I think this is a sensible thing to do anyway, and it fixes a bug with the scroll bars not showing up when a widget is consumed by `napari`.

### DIFF
--- a/magicgui/backends/_qtpy/widgets.py
+++ b/magicgui/backends/_qtpy/widgets.py
@@ -84,6 +84,9 @@ class QBaseWidget(_protocols.WidgetProtocol):
     def _mgui_get_native_widget(self) -> QtW.QWidget:
         return self._qwidget
 
+    def _mgui_get_root_native_widget(self) -> QtW.QWidget:
+        return self._qwidget
+
     def _mgui_get_width(self) -> int:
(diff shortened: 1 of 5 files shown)


dstansby:
That all makes sense, and I think you're repeating yourself from the original PR, so sorry about that. So, it seems we should keep `.native` as the widget that has the layout (and document that if it isn't already, I'll check that), and expose another attribute that 3rd party libraries use for embedding. Since the widget with the layout is a child of the scroll area, there's no way around needing to expose the scroll area for embedding for scrolling to work.

dstansby:
I've added a new property, `root_native_widget`, which is designed to be whathever the top widget is and is designed to be embedded by third parties. Not sure if this a good approach, and the API might need bikeshedding, but thought I'd push some code for discussion 😃

tlambert03:
I think i like it!  will have a play shortly... but this seems like a very good solution

tlambert03 (approved):
looks excellent, thank you @dstansby!

PR #440, opened 2022-07-20
feat: Pass parent to backend widget

tlambert03:
This PR will allow passing a parent to any magicgui widget.

@hanjinliu, this requires that custom widget types accept a parent argument and will affect you.  I ran your tests to make sure they pass with this PR, and they do, but you'll see a lot of `FutureWarning` now.  Let me know if you have any thoughts/concerns

in addition to passing `parent=` all the way up the widget chain, this adds a `parent=` argument to `request_values`, to allow a dialog to be associated with some other widget.  parent can be either a magicgui widget or a native backend widget (if a magicgui widget, `widget.native` will be passed to the backend)


magicgui/widgets/_bases/widget.py
```diff
@@ -90,7 +92,22 @@ def __init__(
         _protocols.assert_protocol(widget_type, prot)
         self.__magicgui_app__ = use_app()
         assert self.__magicgui_app__.native
-        self._widget = widget_type(**backend_kwargs)
+        if isinstance(parent, Widget):
+            parent = parent.native
+        try:
+            self._widget = widget_type(parent=parent, **backend_kwargs)
```
tlambert03, line 98: 👀 @hanjinliu  ... this is the specific line that would break magic-class.  There's a fallback to the old behavior with a warning below

hanjinliu:
Hi @tlambert03 ,
thank you for notifying me.
I totally agree with those changes. It's quite reasonable for the backend widgets to accept any keyword inputs!
I'll see and fix my codes.

tlambert03:
@Czaki, would be good to get your review on this one

magicgui/widgets/_bases/widget.py
```diff
@@ -66,6 +67,7 @@ def __init__(
         visible: bool | None = None,
         enabled: bool = True,
         gui_only=False,
+        parent: Any = None,
```
Czaki, line 70: The `parent` is not documented in the docstring. As this is a fragile part of this PR. There should be informed that if the parent widget is removed (for example, Qt native one) then the object will be unavailable, and trying to access it may end with `RuntimeError: C++ wrapped object has been destroyed.` 

I think that it is fragile as many magicgui users are not familiar with Qt and its errors.

magicgui/widgets/_dialogs.py
```diff
@@ -69,6 +70,8 @@ def request_values(
         interpreted as the `annotation` in `create_widget`, by default ()
     title : str
         An optional label to put at the top., by default ""
+    parent : Widget, optional
+        An optional parent widget, by default None
```
Czaki, line 74:
```suggestion
        An optional parent widget, by default None.
        The dialog will inherit style from the parent object.
```
Or something similar

Czaki (commented):
I see a big profit from this PR, but I have worries about exposing users to the Qt cleaning system. The profit from this PR is big enough to implement this. But proper information should be added to the documentation. 

I also think it may be profitable to connect to [destroyed](https://doc.qt.io/qt-6/qobject.html#destroyed) signal to prepare a better error message or reinstance widget on failure.

magicgui/widgets/_bases/widget.py
tlambert03: added docs,  it's too much to tackle in this PR.  An approval would be appreciated so I can merge (without having to force it)

PR #443, opened 2022-08-02
Fix FunctionGui behavior as a class method

### LINKED ISSUE #409: multiple decorated class methods point to the same object
**Describe the bug**

```py
class T:
    @magicgui
    def testB(self, b: str = 'b'):
        print(b)

    @magicgui
    def testA(self, a: str = 'a'):
        print(a)

t = T()
print(t.testA is t.testB)  # True
```

hanjinliu:
Fixes #409 

Will this solve all the problems of concern?


tlambert03 (approved):
ahh, rookie mistake :joy:   thanks @hanjinliu!

PR #446, opened 2022-08-23
GUIModel: new dataclass-based API, support pydantic and python dataclasses

tlambert03:
this will supersede #379 and #318 
docs to follow

```python
from magicgui import GUIModel, Field
from datetime import date

class MyModel(GUIModel):
    name: str
    age: int
    married: bool
    salary: float = Field(..., ge=0, le=999999)
    birthday: date

obj = MyModel(
    name="Frank", age=55, married=True, salary=100000, birthday=date(1980, 1, 1)
)
obj.gui.show(run=True)

obj.age = 42
assert obj.age == obj.gui.age.value == 42
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.37.2
+    rev: v2.37.3
     hooks:
       - id: pyupgrade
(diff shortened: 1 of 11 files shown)


tlambert03:
closing this in favor of #475  ...  this PR was going to add pydantic as a dependency, and require using `magicgui.GuiModel` as a base class.  I think we implement something lighter that will also support native dataclasses, attrs, NamedTuple, TypedDict, etc...  see #474 for discussion

PR #447, opened 2022-08-24
add test for magic-class

tlambert03:
in addition to testing napari ...

since magic-class is a main user of magicgui, this gives us additional warning if we're breaking something for downstream devs

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -91,6 +91,31 @@ jobs:
         with:
           run: pytest test_napari.py -v --color=yes
 
+  test_magicclass:
+    name: magic-class tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        with:
+          repository: hanjinliu/magic-class
+          path: magic-class
+      - uses: tlambert03/setup-qt-libs@v1
+      - uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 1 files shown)


hanjinliu:
Hi, @tlambert03 ,
very sorry for the late reply. I was on a holiday trip...

It looks like you have found the dependencies, but I want to add that the dependencies of `pyqtgraph` and `vispy` are kind of extra ones. They are not necessary unless users want to import some of the widgets that are dependent of these modules. Widgets that dependent on these modules do not use the non-public API of magicgui.

In my tests their testing directories are in a different places. Maybe I should somehow fix the configuration in magic-class to avoid magicgui's test being too heavy. What do you think?

tlambert03:
> I want to add that the dependencies of pyqtgraph and vispy are kind of extra ones. They are not necessary unless users want to import some of the widgets that are dependent of these modules

Yep, but they're necessary to run "pytest" (without ignoring specific tests), so they fall under your testing dependencies. 

You could either setup a `testing` extra (see the setup.cfg here for an example), or use pytest.import orskip  to make sure your tests run even if they aren't present.

It doesn't make magicgui tests any heavier, it's just one "additional " test, on top of the rest.  Kinda like we test napari too.  (Have a look at the workflow file here if you're curious how I'm testing).

In any case, it's fine either way :). I'll keep testing it as long as it continues to be useful :)

hanjinliu:
I see, thanks for the info!
I'll try to change the testing in a few days.

PR #458, opened 2022-10-17
Support `Literal` type

hanjinliu:
This PR adds typing supports for `Literal[...]` (as `ComboBox`) and `set[Literal[...]]` (as `Select`).
`Select` widget for the ipynb backend was also implemented in a simplest way just for the testing.


tests/test_widgets.py
```diff
@@ -1071,3 +1071,25 @@ def _exec(self, **k):
...
+    assert type(cbox) is widgets.ComboBox
+    assert cbox.choices == ("a", "b", "c")
+
+    @magicgui
+    def f(x: Set[Literal["a", "b", "c"]]):
+        ...
+
+    sel = f.x
+    assert type(sel) is widgets.Select
+    assert cbox.choices == ("a", "b", "c")
```
tlambert03, line 1095:
```suggestion
    from typing_extensions import Literal, get_args

    Lit = Literal["a", 1, True, None, b'bytes']

    @magicgui
    def f(x: Lit):
        ...

    cbox = f.x
    assert type(cbox) is widgets.ComboBox
    assert cbox.choices == get_args(Lit)

    @magicgui
    def f(x: Set[Lit]):
        ...

    sel = f.x
    assert type(sel) is widgets.Select
    assert sel.choices == get_args(Lit)
```

let's test all the literal types, just for completeness

tlambert03 (approved):
thanks!

PR #464, opened 2022-10-22
Fix ipynb PushButton

### LINKED ISSUE #463: PushButton clicked event is not triggered
**Describe the bug**

In ipynb backend, clicking a PushButton does not trigger `changed` event. Callback functions are correctly triggered if the signal is programmatically emitted.

https://user-images.githubusercontent.com/40591297/197326175-f00d404f-6d48-45e9-a6c5-117f0f2e5df4.mp4

**To Reproduce**

Run following code in jupyter notebook or jupyter lab, and click the button.

```python
from magicgui import magicgui, use_app
from magicgui.widgets import PushButton, Container, LineEdit

use_app("ipynb")

btn = PushButton(text="click")
line = LineEdit()

@btn.clicked.connect
def f():
    line.value = "a"

c = Container(widgets=[btn, line])
c.show(True)
```

**Expected behavior**
Event should be triggered.

**Screenshots**

 - OS: Windows 11
 - backend: ipywidgets=8.0.2
 - magicgui version 0.6.0rc0

hanjinliu:
Fixes #463 .

The reason was that `ValueWidget._on_value_changed` does not emit signal with the default argument `value=None`.

### DIFF
--- a/magicgui/backends/_ipynb/widgets.py
+++ b/magicgui/backends/_ipynb/widgets.py
@@ -322,7 +322,14 @@ class PushButton(_IPyButtonWidget):
     _ipywidget: ipywdg.Button
 
     def _mgui_bind_change_callback(self, callback):
-        self._ipywidget.on_click(lambda e: callback())
+        self._ipywidget.on_click(lambda e: callback(False))
+
+    # ipywdg.Button does not have any value. Return False for compatibility with Qt.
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
Thanks for catching this!  Will be fun to have a little more notebook usage :)

Issue #461, opened 2022-10-21
On macOS python 3.10 floats are not being cast to ints.


psobolewskiPhD:
**Describe the bug**
Originally noticed by failing tests in `stardist-napari`:
https://github.com/stardist/stardist-napari/actions/runs/3296152380/jobs/5435481050
But I've verified this in person as well. Both cases it's pyqt5 5.15.7 (both CI and local)
I think this is a python 3.10 thing though based on : https://bugs.python.org/issue37999
But, with pyside6 6.3.2, this doesn't occur, but with pyqt6 6.4.0 it does occur. I can't test with pyside2, because it's broken on arm64. So I'm not sure.
Edit: not sure how I didn't put the traceback in, sorry!
```
File ~/Dev/miniforge3/envs/stardist-310/lib/python3.10/site-packages/magicgui/widgets/_bases/container_widget.py:200, in ContainerWidget._unify_label_widths(self=<FunctionGui plugin(viewer: napari.viewer.Viewer...Future[typing.List[napari.types.LayerDataTuple]]>)
    198 labeled_widget = w._labeled_widget()
    199 if labeled_widget:
--> 200     labeled_widget.label_width = widest_label
        widest_label = 183.25
        labeled_widget = Labeled<EmptyWidget (name='viewer')>

File ~/Dev/miniforge3/envs/stardist-310/lib/python3.10/site-packages/magicgui/widgets/_bases/container_widget.py:100, in ContainerWidget.__setattr__(self=Labeled<EmptyWidget (name='viewer')>, name='label_width', value=183.25)
     94         if name == widget.name:
     95             raise AttributeError(
     96                 "Cannot set attribute with same name as a widget\n"
     97                 "If you are trying to change the value of a widget, use: "
     98                 f"`{self.__class__.__name__}.{name}.value = {value}`",
     99             )
--> 100 object.__setattr__(self, name, value)
        self = Labeled<EmptyWidget (name='viewer')>
        name = 'label_width'
        value = 183.25

File ~/Dev/miniforge3/envs/stardist-310/lib/python3.10/site-packages/magicgui/widgets/_concrete.py:1024, in _LabeledWidget.label_width(self=Labeled<EmptyWidget (name='viewer')>, width=183.25)
   1022 @label_width.setter
   1023 def label_width(self, width):
-> 1024     self._label_widget.min_width = width
        width = 183.25
        self._label_widget = Label(value='viewer', annotation=None, name='')
[...]

PR #465, opened 2022-10-22
use math.ceil to ensure get_text_width returns int

psobolewskiPhD:
Fixes https://github.com/napari/magicgui/issues/461

I'm not sure this is strictly the best way of doing it, but this function is the source of the issue, because it can return a float as the width of the formatted text.
By using `math.ceil` the width value becomes an int and ensures the width will be sufficient (rounding up).

Alternately, one could use `int()` or use `int()` in the downstream functions that use `width`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,49 @@
 # Changelog
 
-## [0.5.1](https://github.com/napari/magicgui/tree/0.5.1) (2022-06-14)
+## [v0.6.0](https://github.com/napari/magicgui/tree/v0.6.0) (2022-10-21)
 
-[Full Changelog](https://github.com/napari/magicgui/compare/v0.5.0...0.5.1)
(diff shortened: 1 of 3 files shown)


tlambert03:
thanks @psobolewskiPhD! 

I like this change, (particularly since it fixes an uncaught typing error, given that the function annotation states that it returns an `int`)...
_and_ I think we should also go one step further and catch the case of a user setting the value to a float, by directly casting to int on line ~109:  `self._qwidget.setMinimumWidth(int(value))` and line ~135 `self._qwidget.setMinimumHeight(int(value))`
can you add those two additional changes?

psobolewskiPhD:
@tlambert03 Yup! I actually did that first, but then thought this was better, but you're right that there is the other fail possibility.

tlambert03 (approved):
thanks @psobolewskiPhD!

PR #470, opened 2022-10-22
feat: add context manager for register_type

### LINKED ISSUE #466: Feature: add context manager to temporarily change the widget associated with a type
@Czaki had a good request:  a context manager like the following:

```python
with type_registered(MyType, widget_type=MyWdgClass):
    ...
``` 

that would restore any previously existing association with `MyType` when the context is exited

tlambert03:
closes #466 

@Czaki ... if you have a moment to review, it would be appreciated!

### DIFF
--- a/magicgui/type_map.py
+++ b/magicgui/type_map.py
@@ -7,8 +7,19 @@ import pathlib
 import types
 import warnings
 from collections import defaultdict
+from contextlib import contextmanager
 from enum import EnumMeta
-from typing import Any, Callable, DefaultDict, ForwardRef, Type, TypeVar, cast, overload
+from typing import (
+    Any,
+    Callable,
+    DefaultDict,
+    ForwardRef,
+    Iterator,
+    Type,
+    TypeVar,
+    cast,
+    overload,
+)
 
 from typing_extensions import Literal, get_origin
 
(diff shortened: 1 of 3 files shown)


magicgui/type_map.py
```diff
@@ -369,6 +380,58 @@ def _deco(type_):
...
+    null = object()
+
+    prev_return_callback = _RETURN_CALLBACKS.get(_type_, null)
+    prev_type_def = _TYPE_DEFS.get(_type_, null)
+    try:
+        register_type(
+            type_, widget_type=widget_type, return_callback=return_callback, **options
+        )
+        yield
+    finally:
```
Czaki, line 424: what did you think about add here check if values under `_RETURN_CALLBACKS[_type_]` and `_TYPE_DEFS[_type_]`  are such as expected and if not (someone calls `register_type` for such type inside context manager) then print warning message?

magicgui/type_map.py
```diff
@@ -369,6 +380,58 @@ def _deco(type_):
...
+    tw.resolve()
+    _type_ = tw.outer_type_
+    null = object()
+
+    prev_return_callback = _RETURN_CALLBACKS.get(_type_, null)
+    prev_type_def = _TYPE_DEFS.get(_type_, null)
+    try:
+        register_type(
+            type_, widget_type=widget_type, return_callback=return_callback, **options
+        )
```
Czaki, line 422: having this under `try` clause is to keep the consistency of the data structure. If yes, then maybe this should be done in `register_type` itself?

magicgui/type_map.py
```diff
@@ -369,6 +380,58 @@ def _deco(type_):
...
+        yield
+    finally:
+        if prev_return_callback is null:
+            _RETURN_CALLBACKS.pop(_type_, None)
+        else:
+            _RETURN_CALLBACKS[_type_] = prev_return_callback  # type: ignore
+        if prev_type_def is null:
+            _TYPE_DEFS.pop(_type_, None)
+        else:
+            _TYPE_DEFS[_type_] = prev_type_def  # type: ignore
```
Czaki, line 432: looking at test and coverage messages, it looks like something is wrong. Or are default widgets stored outside `_TYPE_DEFS` dict?

tlambert03: default widgets are stored outside of `_TYPE_DEFS` (for better or worse)

magicgui/type_map.py
tlambert03: i agree, bit odd in the try block there.  gonna just move it above.

magicgui/type_map.py
tlambert03: added, thanks

tlambert03:
many thanks for the review.  updated

magicgui/type_map.py
```diff
@@ -369,6 +381,60 @@ def _deco(type_):
...
+    _type_ = register_type(
+        type_, widget_type=widget_type, return_callback=return_callback, **options
+    )
+    new_type_def: Optional[WidgetTuple] = _TYPE_DEFS.get(_type_, None)
+    try:
+        yield
+    finally:
+
+        if return_callback is not None:
+            _RETURN_CALLBACKS[_type_].remove(return_callback)
```
Czaki, line 427: in the previous implementation, you restore also value of `_RETURN_CALLBACKS` here you only remove some value. 

What happen if someone use `register_type` with both `widget_type` and `return_callback` inside? 

Maybe also callback should be removed only if is not present in `_RETURN_CALLBACKS[_type_]` before the context manager usage?

tlambert03: > Maybe also callback should be removed only if is not present in _RETURN_CALLBACKS[_type_] before the context manager usage?

sounds good, added

PR #475, opened 2022-10-23
feature: add `UiField` as the new intermediate object representing widget parameters

tlambert03:
see #474 for the complete context and motivation for this PR

`UiField` will be the thing that all type annotations are coerced to prior to widget generation (think of it as a UI-focused variant of `dataclasses.field`).

this is an alternative to #446 ... it is less "heavy" in that it doesn't bring in the pydantic dependency, instead, aiming to support pydantic, attrs, dataclasses, named tuples, etc... 

All of these are a set of `name: annotation = 'defaut_value'` .... which can be reduced to a "model" that is a collection of field objects.  This PR adds that field object, and adds some tests that begin to demonstrate intended usage.

everything is private still, so this will be just one of a number of PRs moving in this direction


tlambert03:
just some notes for myself about the mapping between any current widget parameter names and corresponding fields in UiField (if applicable):

```python

# these params directly map to the UiField names
{
    "name",
    "visible",
    "nullable",
    "orientation",
}

# aliases
{
    "annotation": "type",
    "choices": "enum",  # CategoricalWidget
    "enabled": "NOT disabled",
    "label": "title",
    "max": "maximum",
    "min": "minimum",
    "start": "minimum",  # "RangeEdit",
    "step": "multiple_of",  # "RangedWidget" ... complicated, not true for RangeEdit
    "stop": "exclusive_maximum",  # "RangeEdit",
    "text": "title", # ButtonWidgets have "text"... confusing
    "tooltip": "description",
    # 'value': 'default',
}

# no correlate:
{
    "gui_only": "RadioButtons",
    "parent": "RadioButtons",
    "backend_kwargs": "RadioButtons",
    # used for options of child widgets ... but they should be done with annotations?
    "options": "ListEdit",
    "min_pos": "TransformedRangedWidget",
    "max_pos": "TransformedRangedWidget",
    "bind": "RadioButtons",
    "base": "LogSlider",
    "tracking": "SliderWidget",
    "readout": "SliderWidget",  # toggles editable spinbox visibility.
    # allow_multiple is used to induce a Select widget...
    # but it is most like an array type with items:
    # {
    # "type": "array",
    # "items": {
    #     "type": "string",
    #     "enum": [ "foo", "bar", "baz" ]
    # },
    # "uniqueItems": true,
    # "minItems": 1
    # }
    # DEPRECATE
    "allow_multiple": "CategoricalWidget",
    "mode": "FileEdit",
    "filter": "FileEdit",  # most like 'pattern'??
}
# Table stuff (don't map these... can be done directly with value)
{
    "index": "Table",
    "columns": "Table",
}
# container stuff
{
    "app": "FunctionGui",
    "auto_call": "FunctionGui",
    "result_widget": "FunctionGui",
    "param_options": "FunctionGui",
    "persist": "FunctionGui",
    "raise_on_unknown": "FunctionGui",
    "layout": "Container",
    "scrollable": "Container",
    "widgets": "Container",
    "labels": "Container",
    "function": "MainFunctionGui",
    "call_button": "FunctionGui",
    "tooltips": "FunctionGui",
}
```

PR #476, opened 2022-10-23
Enable option to raise excpetion if magicgui cannot determine widget for provided value/annotation

### LINKED ISSUE #473: Verbose fail on lack of matching widget for provided type
When working with magicgui one of the problems is that it silently defaults to `widgets.EmptyWidget` for an unknown type. So it means that there is no clear way to verify if created widget contains all required fields without manually type-checking of each field. Writing such a test is really time-consuming. 

It will be nice to have one of these options:

1. change behavior to raise an exception (for example, ValueError) instead of return EmptyWidget:
   https://github.com/napari/magicgui/blob/9e29739b4047dd2fb7b223152513170aad4a1ebc/magicgui/type_map.py#L211
2. allow control of this behavior from code. This allows to at least tests to get an exception instead of a silent fail of matching types. 
3. same as above but with the default behavior change to raise an exception. 

I could implement it, but I would ask you which option you prefer @tlambert03.

Czaki:
This PR adds the option to raise exceptions if magicgui cannot determine a widget or a given type. 

By default, the exception is not raised if using `magicgui` decorator but is raised if `create_widget` is called directly. Is it OK for you @tlambert03 

closes #473

### DIFF
--- a/magicgui/_magicgui.py
+++ b/magicgui/_magicgui.py
@@ -25,6 +25,7 @@ def magicgui(
     main_window: bool = False,
     app: AppRef = None,
     persist: bool = False,
+    raise_on_unknown: bool = False,
     **param_options: dict,
 ):
     """Return a :class:`FunctionGui` for ``function``.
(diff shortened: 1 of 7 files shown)


tlambert03:
yeah, I think this looks good to me.  It's a shame this has to go all the way through `signature.py` and everything... but I took a look and I agree it seems to be the way to do it.  (would probably be a bit easier with the pattern proposed in #474 ... but we can just make sure to respect the same intention as whatever you implement here when that is implemented)

let me know when this is ready for final review (could be now?)

Czaki:
> let me know when this is ready for final review (could be now?)

there is no test (i will add) but you could review everything else.

tlambert03 (approved):

Czaki:
ready

tlambert03 (approved):
thanks!

PR #497, opened 2022-11-12
refactor: repo reorganization and return removed names

tlambert03:
The goal of this reorganization is to make clearer top-level modules that delineate the main components of magicgui

Screen Shot 2022-11-13 at 8 55 02 AM

There are a lot of changes here, but I have tested both napari and magic-class to make sure that everything still passes (albeit with some warnings), and opened https://github.com/napari/napari/pull/5331 to remove one of those warnings

This also returns some of the names removed in #448 and #475 (to get magic-class tests working again)


tlambert03:
tests will pass here after https://github.com/napari/napari/pull/5331

PR #499, opened 2022-11-13
More reasonable signal emission of `ListEdit` and `TupleEdit`

hanjinliu:
I found that `ListEdit` and `TupleEdit` are emitting `self` now, and probably in wrong number of times (which may cause excessive number of function calls). This PR fixes these problems.

This change might break backward compatibility if one uses `list_widget.changed.connect`. Since they are supposed to be used as value-like widgets, they should emit their values just like `FileEdit`, so this change should be applied eventually.


magicgui/widgets/_concrete.py
```diff
@@ -749,28 +750,43 @@ def annotation(self, value):
         self._annotation = value
         self._args_type = arg
 
+    def __iter__(self) -> Iterator[ValueWidget]:
```
tlambert03, line 753: Container is now Generic as of #475 ... so these aren't necessary anymore if we declare `class ListEdit(Container[ValueWidget])` ... but I'll do that in a followup PR

tests/test_widgets.py
```diff
@@ -968,29 +968,56 @@ def test_list_edit():
...
+    list_edit.changed.connect(mock)
     assert list_edit.value == [1, 2, 3]
     assert list_edit.data == [1, 2, 3]
+    assert mock.call_count == 0
 
     list_edit.btn_plus.changed()
     assert list_edit.value == [1, 2, 3, 3]
     assert list_edit.data == [1, 2, 3, 3]
+    assert mock.call_count == 1
+    mock.assert_called_with([1, 2, 3, 3])
```
tlambert03, line 983: 👍

tlambert03 (commented):
thanks @hanjinliu!  Merging and will follow up on minor typing suggestions in another PR

tlambert03 (approved):

hanjinliu:
Thank you!

PR #513, opened 2022-11-23
fix bug in show_file_dialog

MBPhys:
Hi guys,

first of all fantastic work to make python applications graphical user interfacing more and more.

I have used this to make my own GUI applications. I noticed that there is a very small bug in the show_file_dialog function.
It does not work for the directory mode, because the if condition is not correct. I fix this.

Image_bug


tlambert03 (approved):
thank you for the fix!

PR #515, opened 2022-11-25
feat: add password field

### LINKED ISSUE #514: Password field
Hi Talley @tlambert03 , hi all,

I _think_ at the moment it is not possible to have password fields using magicgui ([see related image.sc thread](https://forum.image.sc/t/magicgui-password-field/74333/2)). It would be cool mark a field as password field, e.g. in such a function:

```
def login(server:str, username:str, password:str):
    pass
```

Thanks!

Best,
Robert

tlambert03:
closes #514 by adding a new `widgets.Password`

You can get a password widget, as usual, by specifying `widget_type` of `'Password'`... but I also special cased arguments named "password" with an annotation of "str" that do _not_ declare a `widget_type`.

both `password` and `password2` will be Password widgets here:

```python
@magicgui(password2={"widget_type": "Password"})
def login(username: str, password: str, password2: str):
    ...
```

Screen Shot 2022-11-25 at 12 28 22 PM


haesleinhuepf:
Amazing. Thanks so much! 🎉

PR #524, opened 2022-12-08
Add directory picker example to file_dialog.py

GenevieveBuckley:
I thought it would be helpful to add a directory picker example to examples/file_dialog.py
People might be unaware that's a thing you can do, so it would be good to highlight it as well.


tlambert03 (approved):
Thanks!

PR #536, opened 2023-02-24
Support nested `Annotated` types

### LINKED ISSUE #535: Cannot use nested `Annotated` type
**Describe the bug**
The recent update in `make_annotated` makes it unable to interpret nested `Annotated` types (although I'm not sure if "magicgui.signature" will still be used).

**To Reproduce**

Run this code.

```python
from typing import Annotated
from magicgui import magicgui
@magicgui
def f(x: Annotated[Annotated[int, {"max": 5}], {"min": 0}]):
    print(x)
```

**Expected behavior**
`Annotated[Annotated[int, {"max": 5}], {"min": 0}]` should be interpreted as `Annotated[int, {"max": 5, "min": 0}]`

**Screenshots**

 - OS: Windows 11
 - backend: PyQt5
 - magicgui version 0.7.0rc1

hanjinliu:
Fixes #535 .

`get_args` returns multiple options if `Annotated` is nested. We can also easily support annotated-types in the future because it also uses multiple options like `Annotated[int, Gt(3), Lt(10)]` (raises a TypeError for now).

### DIFF
--- a/src/magicgui/signature.py
+++ b/src/magicgui/signature.py
@@ -54,8 +54,10 @@ def make_annotated(annotation: Any = Any, options: dict | None = None) -> Any:
     _options = (options or {}).copy()
 
     if get_origin(annotation) is Annotated:
-        annotation, anno_options = get_args(annotation)
-        _options.update(anno_options)
+        # nested Annotated type results in multiple dicts.
+        annotation, *anno_options = get_args(annotation)
+        for opt in anno_options:
+            _options.update(opt)
     return Annotated[annotation, _options]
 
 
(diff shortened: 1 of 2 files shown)


src/magicgui/signature.py
```diff
@@ -54,8 +54,12 @@ def make_annotated(annotation: Any = Any, options: dict | None = None) -> Any:
     _options = (options or {}).copy()
 
     if get_origin(annotation) is Annotated:
-        annotation, anno_options = get_args(annotation)
-        _options.update(anno_options)
+        # nested Annotated type results in multiple dicts.
+        annotation, *anno_options = get_args(annotation)
+        for opt in anno_options:
+            if not isinstance(opt, dict):
```
tlambert03, line 60: hmm... was I doing this when it worked before?  I'm not sure I want to prevent people from using Annotated for any other reason.

tlambert03: Eh, I guess I'm already preventing it 😂 let's go with it!

tlambert03 (approved):

PR #539, opened 2023-03-05
Fix recursion error when connecting to container self.method during subclass init 

### LINKED ISSUE #538: `RecursionError` with `psygnal==0.8.1`
**Describe the bug**
It seems that the latest update of the weak callbacks in `psygnal` sometimes causes recursion error in magicgui widgets.

**To Reproduce**

Following code raises RecursionError (Python<3.11 crashes silently).

```python
from magicgui.widgets import Container, PushButton

class C(Container):
    def __init__(self):
        btn = PushButton()
        btn.changed.connect(self._on_clicked)
        super().__init__(widgets=[btn])
    def _on_clicked(self):
        print(0)
c = C()
```

Connecting callback after construction does not.

```python
from magicgui.widgets import Container, PushButton

class C(Container):
    def __init__(self):
        btn = PushButton()
        super().__init__(widgets=[btn])
        btn.changed.connect(self._on_clicked)
    def _on_clicked(self):
        print(0)
c = C()
```

I can walkaround this by passing non-method type, such as lambda function.

```python
from magicgui.widgets import Container, PushButton

class C(Container):
    def __init__(self):
        btn = PushButton()
        btn.changed.connect(lambda: self._on_clicked())
        super().__init__(widgets=[btn])
    def _on_clicked(self):
        print(0)
c = C()
```

**Expected behavior**

**Screenshots**

 - OS: Windows 11
 - backend: PyQt5
 - magicgui version 0.7.0

tlambert03:
fixes #538  (the issue is that `self._list` wasn't yet assigned when calling `Container.__getattr__` from the super-init during construction... rest of changes are just moving container tests to a new file).

This was definitely a magicgui bug and not a psygnal bug, since you essentially couldn't call `getattr` on the container during subclass initialization

thanks for reporting @hanjinliu!  will get a new release out shortly

### DIFF
--- a/src/magicgui/widgets/bases/_container_widget.py
+++ b/src/magicgui/widgets/bases/_container_widget.py
@@ -81,6 +81,7 @@ class ContainerWidget(Widget, _OrientationMixin, MutableSequence[WidgetVar]):
     )
     _widget: protocols.ContainerProtocol
     _initialized = False
+    _list: list[WidgetVar] = []
 
     def __init__(
         self,
@@ -107,7 +108,7 @@ class ContainerWidget(Widget, _OrientationMixin, MutableSequence[WidgetVar]):
 
     def __getattr__(self, name: str) -> WidgetVar:
         """Return attribute ``name``.  Will return a widget if present."""
-        for widget in self:
+        for widget in self._list:
             if name == widget.name:
                 return widget
         return object.__getattribute__(self, name)  # type: ignore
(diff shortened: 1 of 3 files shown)


src/magicgui/widgets/bases/_container_widget.py
```diff
@@ -81,6 +81,7 @@ class ContainerWidget(Widget, _OrientationMixin, MutableSequence[WidgetVar]):
     )
     _widget: protocols.ContainerProtocol
     _initialized = False
+    _list: list[WidgetVar] = []
```
tlambert03, line 84: this line is the only critical line in this PR

Issue #541, opened 2023-03-06
After update to 0.7.0 magicgui stop handling optional return


Czaki:
**Describe the bug**

Up to version 0.6.1, it is possible to annotate `Optional[ImageData]`, and it works without problems. 

Since 0.7.0 the 
https://github.com/pyapp-kit/magicgui/blob/b675ca9a3d0c24629730a92947a91799816e10e0/src/magicgui/widgets/_function_gui.py#L342
does not return `add_layer` callback

Is this a real bug change or an intentional change that needs to be handled on the napari side? 

**To Reproduce**

Use test form this PR with magicgui 0.6.1 and 0.7.2
https://github.com/napari/napari/pull/5595

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. Windows, Mac]
 - backend: [e.g. Qt 5.13.0]
 - magicgui version [e.g. 0.1.0]

tlambert03:
thanks @Czaki, good find.  I'd say that calling a callback that was registered for `Type` when someone annotates with `Optional[Type]` was an "unintended feature", which was never tested explicitly.  It came as a side effect before of vendoring pydantic's type wrapper, which was since removed in #448 

I think [this line](https://github.com/pyapp-kit/magicgui/pull/448/files#diff-00e91ee7ed095b8d66caa3b10aed37139a0b2602c83939a537e7acc2454d6b17L469-L472) here was allowing for `Optional[Type]` to still call the callback.

It's something of a tricky situation I guess, since I wouldn't have knowingly opted to provide implicit support for calling a callback that says it needs a `Thing` when someone says their function returns an `Optional[Thing]` (at the very least, without checking that `return_value is not None`...)

So, we've got a few options now (let me know what you prefer):
1. `napari` can explicitly opt in to supporting `Optional[ImageData]` by registering *both* `ImageData` and `Optional[ImageData]` in this line here: https://github.com/napari/napari/blob/c77e001bd7ed21303aba7a547aafdf3c90b1fed6/napari/types.py#L143-L148
2. I can add back in this accidental feature to magicgui, meaning that when anyone uses `register_type(Type, return_callback=my_func)`, then `my_func` *may indeed* be potentially called even without an instance of `Type` ... that feels icky
[...]

PR #544, opened 2023-03-10
Register all types in a return_callback Union

tlambert03:
part of a fix to #541, this makes it so that when you use `register_type(Union[...], return_callback=...)`, then you effectively register the callback for functions that return either the union or any type within the Union (not just the actual union object itself)

### DIFF
--- a/src/magicgui/type_map/_type_map.py
+++ b/src/magicgui/type_map/_type_map.py
@@ -418,10 +418,16 @@ def register_type(
 
     def _deco(type_: _T) -> _T:
         _type_ = resolve_single_type(type_)
-
         if return_callback is not None:
             _validate_return_callback(return_callback)
-            _RETURN_CALLBACKS[_type_].append(return_callback)
+            # if the type is a Union, add the callback to all of the types in the union
+            # (except NoneType)
+            if get_origin(_type_) is Union:
+                for t in get_args(_type_):
+                    if _is_none_type(t):
+                        _RETURN_CALLBACKS[t].append(return_callback)
+            else:
+                _RETURN_CALLBACKS[_type_].append(return_callback)
 
         _options = cast(dict, options)
 
@@ -522,8 +528,9 @@ def type2callback(type_: type) -> list[ReturnCallback]:
     if type_ is inspect.Parameter.empty:
         return []
 
-    # look for direct hits
-    type_ = resolve_single_type(type_)
+    # look for direct hits ...
+    # if it's an Optional, we need to look for the type inside the Optional
+    _, type_ = _is_optional(resolve_single_type(type_))
     if type_ in _RETURN_CALLBACKS:
         return _RETURN_CALLBACKS[type_]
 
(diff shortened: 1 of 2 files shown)


tlambert03:
cc @Czaki, if you have a moment to review, this implements option 3 from https://github.com/pyapp-kit/magicgui/issues/541#issuecomment-1456940458

src/magicgui/type_map/_type_map.py
```diff
@@ -532,3 +539,12 @@ def type2callback(type_: type) -> list[ReturnCallback]:
         if safe_issubclass(type_, registered_type):
             return _RETURN_CALLBACKS[registered_type]
     return []
+
+
+def _is_optional(type_: Any) -> tuple[bool, type]:
```
Czaki, line 544: from documentation optional is equivalent of `Union[X, None]` with order. Did you meet in a different order, or did you write this in this way to guarantee safety?

tlambert03: it's true that `Optional[X]` results in `Union[X, None]` ... but it's not the only way to create an optional in later python versions

```python
In [2]: get_args(int | None)
Out[2]: (<class 'int'>, <class 'NoneType'>)

In [3]: get_args(None | int)
Out[3]: (<class 'NoneType'>, <class 'int'>)
```

src/magicgui/type_map/_type_map.py
```diff
@@ -418,10 +418,16 @@ def register_type(
 
     def _deco(type_: _T) -> _T:
         _type_ = resolve_single_type(type_)
-
         if return_callback is not None:
             _validate_return_callback(return_callback)
-            _RETURN_CALLBACKS[_type_].append(return_callback)
+            # if the type is a Union, add the callback to all of the types in the union
+            # (except NoneType)
+            if get_origin(_type_) is Union:
```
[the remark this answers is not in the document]
tlambert03: fixed

tlambert03:
> So I understand that After merging it there should be added if to napari codebase to check it depends on the magicgui version?

yeah, I can do that PR if you'd like?

tlambert03:
updated the docs a bit.  thanks for the review @Czaki !

tlambert03:
> But code in magicgui does not contain deduplication of callbacks.

this is a good point... but i *think* in most cases it's going to be fine (the second one will overwrite the first one).  let me double check that for you

Czaki:
> this is a good point... but i _think_ in most cases it's going to be fine (the second one will overwrite the first one). let me double check that for you

Yes and the whole recalculation of all attributes and double transfer to GPU memory of layer data?

tlambert03:
> Yes and the whole recalculation of all attributes and double transfer to GPU memory of layer data?

no, I don't mean the second "layer" will overwrite the first "layer", I mean the second registered callback will overwrite the first registered callback (in the type2callback map) and only a single one will actually be called at runteim

tlambert03:
buuut nope... it gets called twice.  will fix

Issue #543, opened 2023-03-07
Consider deprecating return_callback feature of register_type


tlambert03:
just thinking aloud here as I think about this topic (and will probably make a new issue for this).  In retrospect, I think the `return_callback` option to `register_type` was probably an overreach of scope for magicgui in the first place.  While it definitely allows some cool behavior in napari, this sort of magical processing of the return value seems better suited to in-n-out, or perhaps implemented directly in napari for this specific use case of adding something to the viewer with a callback connection:

```python
# napari instantiates the widget somewhere
wdg = some_factory()

# connect callback:
if satisfies_some_condition(wdg.return_annotation):
    wdg.called.connect(_mgui.add_layer_data_to_viewer)

# add it to the viewer
viewer.window.add_dock_widget(wdg)
```

that will definitely require a gradual deprecation, but as I think about the complexities of processing return annotation types (e.g. do we support subtypes? etc...) i think it's probably better to remove that feature rather than elaborate upon it

_Originally posted by @tlambert03 in https://github.com/pyapp-kit/magicgui/issues/541#issuecomment-1458227189_

tlambert03:
> I have strong worries that many people outside napari may depend on this magic

yep, if this happens at all, it absolutely won't happen "quickly".  It would be a multi-version deprecation warning, with a clear alternative.  From a quick search, I can only find `napari` and `magic-class` using the `return_callback` argument to `register_type`.  

cc @hanjinliu

PR #545, opened 2023-03-11
fix: prevent dupe calls

tlambert03:
trying to fix the duplicate call thing mentioned in https://github.com/pyapp-kit/magicgui/pull/544#pullrequestreview-1335180278

### DIFF
--- a/src/magicgui/type_map/_type_map.py
+++ b/src/magicgui/type_map/_type_map.py
@@ -50,7 +50,9 @@ class MissingWidget(RuntimeError):
     """Raised when a backend widget cannot be found."""
 
 
-_RETURN_CALLBACKS: DefaultDict[type, list[ReturnCallback]] = defaultdict(list)
+_RETURN_CALLBACKS: DefaultDict[frozenset[type], list[ReturnCallback]] = defaultdict(
+    list
+)
 _TYPE_DEFS: dict[type, WidgetTuple] = {}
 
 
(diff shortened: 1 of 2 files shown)


tlambert03:
ugh... this is just a real annoying problem. 

1. in order to truly support the proper `Optional[type]` return annotation that @Czaki was going for in https://github.com/napari/napari/pull/5595, magicgui needs to move from a simple  `type -> list[callback]` map, to a much more complicated `set[type] -> list[callback]` map... and then when looking up a callback for a return type annotation, look for the first `set[type]` that is a superset of the return callback set.
2. that more complicated map makes everything more complicated:
    - it slows down callback lookup at runtime
    - it makes it much harder to undo type registration (e.g. in the `type_registered` context manager)

basically, this all just further emphasizes that the `return_callback` feature of register_type is far more complicated to do "correctly", and makes me want to deprecate the whole feature altogether as mentioned in #543 ...

I really don't know the best solution here.

tlambert03:
one possibility here is just to temporarily go back to the previous behavior (that I absolutely thing was *incorrect* behavior) of calling callbacks registered for `Type` even when the return annotation of the `@magicgui` function is `Optional[Type]` (as mentioned in https://github.com/pyapp-kit/magicgui/issues/541).

That's a big bummer, cause it's just obviously wrong behavior.  @Czaki, what do you think here?

Czaki:
Why not just check presence of callback in lust before append? Registration type should be rare operation and such list should be short.

tlambert03:
Because if we're dealing with sets as keys you now have to check all keys for a subset match, remember which ones matched, etc...

tlambert03:
It's not that it's not possible, it's just getting uglier and uglier for something that was an accidental "feature/bug" to begin with

Czaki:
Because of `type2callback` return first callback list where superset is found this test will fail:

```
def test_proper_call():
    @magicgui
    def func(a: int, b: int) -> int:
        return a + b

    mock1 = Mock()
    mock2 = Mock()

    register_type(int, return_callback=mock1)
    register_type(Optional[int], return_callback=mock2)

    func(1, 2)

    mock2.assert_called_once()
    mock1.assert_called_once()
```

tlambert03:
closing, see https://github.com/pyapp-kit/magicgui/pull/546

PR #546, opened 2023-03-13
fix: prevent dupe calls, alternative

Czaki:
This is an alternative to #545 implementation to prevent duplicated calls of callbacks.


Czaki:
@tlambert03 What did you think about this option?

src/magicgui/type_map/_type_map.py
```diff
@@ -550,10 +556,8 @@ def type2callback(type_: type) -> list[ReturnCallback]:
...
 
 
-def _is_optional(type_: Any) -> tuple[bool, type]:
-    # TODO: this function is too similar to _type_optional above... need to combine
-    if get_origin(type_) is Union:
-        args = get_args(type_)
-        if len(args) == 2 and any(_is_none_type(i) for i in args):
-            return True, next(i for i in args if not _is_none_type(i))
-    return False, type_
+def _generate_union_variants(type_: Any) -> Iterator[type]:
```
tlambert03, line 559: thanks for putting this together @Czaki.

correct me if I'm wrong, but the primary difference between this and #545 is that this one only stores pairs of types, whereas #545 stores full sets?  I assume you're thinking specifically of the case of `Optional[Type]` here right?  (while ignoring the case of Unions longer than length 2?)

Note, it's also still not working for the `type_registered`, which, while possible, is also part of the ugliness that I commented on in https://github.com/pyapp-kit/magicgui/pull/545#issuecomment-1465235311

Can you tell me what you see as the main difference in this PR?

I know there's a solution somewhere that gets the bare minimum behavior working for optional types, but I really think this whole thing is exposing a deeper ugliness that makes me queasy (hence #543)

Czaki: No. `itertools.combinations` return all combinations of the given length. And the outer loop iterates from 2 to number of union arguments. So it contains all subset of length 2, 3, 4 up to the maximum length. 

The main difference between this two PR is that in this PR all subset of types is generated during registration callback. So, by the cost of memory, triggering callback is faster. In #545 you register a callback in one place, but every time there is a check if there is a callback, the subset existence is checked. 

Also, this PR does not need callback merge and deduplication (see https://github.com/pyapp-kit/magicgui/pull/545#issuecomment-1484867765)

I have fixed `type_registered`

tlambert03: aha!  thanks for clarifying :)  will review more soon.  thanks for fixing the final test

PR #550, opened 2023-04-03
Change exception to not confuse static code checkers

Czaki:
When using `NotImplementedError`, the static code checkers like pylint marks a method as required to implement in a subclass. I suggest changing to `RuntimeError`  (but I'm not glued to this selection of exception class). 

obraz

### DIFF
--- a/src/magicgui/widgets/bases/_container_widget.py
+++ b/src/magicgui/widgets/bases/_container_widget.py
@@ -184,7 +184,7 @@ class ContainerWidget(Widget, _OrientationMixin, MutableSequence[WidgetVar]):
 
     def __setitem__(self, key: Any, value: Any) -> NoReturn:
         """Prevent assignment by index."""
-        raise NotImplementedError("magicgui.Container does not support item setting.")
+        raise RuntimeError("magicgui.Container does not support item setting.")
 
     def __dir__(self) -> list[str]:
         """Add subwidget names to the dir() call for this widget."""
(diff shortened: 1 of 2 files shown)


tlambert03:
oh how interesting and vaguely annoying :joy:

I'm fine with this in this case, `RuntimeError` actually seems like a better match here, but I'd also like to learn more about it.  Can you give me whatever config you're using in pylint? (if that's what's raising the error?)  I'm unable to trigger the error locally with:


```python
# some_file.py

"""Module docstring."""


class SearchLabel(Container):
    """Some Widget."""

    def __init__(self) -> None:
        super().__init__(layout='vertical')
```

```
pylint some_file.py
```

tlambert03:
going to merge, since I have no issues with this.  But if you can give me a config to reproduce that'd be great

Czaki:
hm. The screenshot comes from deepsource https://deepsource.io/gh/4DNucleome/PartSeg/run/f3260ef2-a55e-44b4-9b58-5cca538e70c8/python/PYL-W0223?listindex=0

I just copied your file, updated pylint to the latest one, and call 

```
$ pylint examples/some_file.py                                                                                                                                                                                                                                                                           
************* Module some_file
examples/some_file.py:7:0: W0223: Method '__setitem__' is abstract in class 'ContainerWidget' but is not overridden in child class 'SearchLabel' (abstract-method)

------------------------------------------------------------------
Your code has been rated at 7.50/10 (previous run: 5.00/10, +2.50)
```

Do you know how to dump configuration?

PR #556, opened 2023-04-28
fix: fix binding of non ValueWidgets in guiclass

### LINKED ISSUE #555: guiclass not updating model `Path` parameters
**Describe the bug**
For a class decorated with the experimental `@guiclass` decorator modifying the value of an attribute decorated as `Path` does not update the value on the model

**To Reproduce**

```python
from pathlib import Path

import napari
from magicgui.experimental import guiclass
from magicgui.widgets import Button


@guiclass
class MyGuiClass:
    a: Path = Path('blabla')


gui = MyGuiClass()

def printer():
    print(gui.a.resolve())

button = Button(text='print path')
button.clicked.connect(printer)
gui.gui.append(button)
gui.gui.show()
napari.run()
```

sorry for the napari, easiest way I know to get an event loop 😆 

**Expected behavior**
Model should be updated

**Screenshots**
image

 - OS: Mac
 - backend: Qt 5.15.6
 - magicgui version 0.7.2

tlambert03:
fixes #555 

the deeper issue is that FileEdit implements a very similar protocol to ValueWidget (has a `changed` event and `value`), but doesn't subclass it... so changes to the widget weren't getting bound to the model.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -101,7 +101,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -e .[testing,pyqt5]
-          pip install numpy superqt vispy pyqtgraph
           python -m pip install ./magic-class[testing]
 
       - name: Test magicclass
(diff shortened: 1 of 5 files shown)


tlambert03:
extended the class docstring, and added [more comments inline](https://github.com/funkelab/motile/pull/31/files#diff-4267a412fbefb8186f6fbb3661285e4edac9307d85d03c0a968199e66bbd2660R93-R104)

alisterburt:
Awesome work - thanks for the fix!

PR #563, opened 2023-06-29
Update `choices` setter for categorical widget to ensure `_default_choices` are updated when a `callable` choices is passed

jamesyan-git:
Noticed that the `choices` setter for categorical widgets doesn't update `_default_choices`, even though `_default_choices` is set when a `callable` is passed on `init`. In our plugin we have to set `choices` to a callable after widget creation, but we've noticed the combo box doesn't get populated correctly until we force some other event change. This PR fixes that and makes sure `reset_choices` uses the callable.

I've added a comment and a test that fails on main and passes with this PR.

Let me know if this makes sense or if there's a better way to achieve the desired behaviour! Thanks!

cc @jni

### DIFF
--- a/src/magicgui/widgets/bases/_categorical_widget.py
+++ b/src/magicgui/widgets/bases/_categorical_widget.py
@@ -137,7 +137,8 @@ class CategoricalWidget(ValueWidget[T]):
             str_func = choices["key"]
         elif not isinstance(choices, EnumMeta) and callable(choices):
             _choices = choices(self)
-
+            # The following line ensures that when setting choices with a callable, the default choices are also updated with that callable
+            self._default_choices = choices
         else:
             _choices = choices
 
--- a/src/magicgui/widgets/bases/_categorical_widget.py
+++ b/src/magicgui/widgets/bases/_categorical_widget.py
@@ -137,7 +137,8 @@ class CategoricalWidget(ValueWidget[T]):
             str_func = choices["key"]
         elif not isinstance(choices, EnumMeta) and callable(choices):
             _choices = choices(self)
-            # The following line ensures that when setting choices with a callable, the default choices are also updated with that callable
+            # ensure that when setting choices with a callable,
+            # the default choices are also updated with that callable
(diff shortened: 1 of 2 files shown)


jni:
@jamesyan-git so, this kinda sorta makes sense, but it doesn't exactly feel right to me (@tlambert03 please correct me if I'm wrong). It would be good to clarify the motivation a bit more. If I can summarise what happens:

- you use `widget_init=my_init_func` to set the widget choices in a `magic_factory`
- This kinda sorta works for a bit but then `widget._post_init()` gets called, which itself calls `self.reset_choices()`, and then your setting of the choices gets discarded.
- This PR fixes it because `reset_choices()` grabs the defaults, which you've now overwritten, so it's happy days.

tbh I must say I don't understand why you would `reset_choices` in the `_post_init`. @tlambert03 what's the reason for this?

Either way, @jamesyan-git on our end, can you do `widget._default_choices = callable` and fix things? If so, I suggest going with that regardless of the outcome here. (@tlambert03 our use case is populating the choices with feature columns from a pandas dataframe in the napari layer metadata, [in case you're curious](https://github.com/jni/skan/pull/201/files#diff-53ef07fd6282e87f05ce8b5947f3756ce7b6b4f9b547f947ebfed4f93d99e17bR33); but we found that we can't get the choices to show up on widget launch.)

DragaDoncila:
I think the `callable` was always meant to be used for `_default_choices`. The first line in the `init` sets `_default_choices=choices` without making any checks on it. I think it makes sense for the setter to also update `_default_choices` when a callable choices is passed. To me looking at this code, it seems like an oversight that it wasn't set as `_default_choices` to begin with - even the `reset_choices` docstring mentions that if `choices` are `callable`, this function may return different things.

tlambert03:
yep, this looks great, thanks for the nice PR @jamesyan-git 👍

Sorry in general for the confusing choices behavior.  it's long been a target of cleanup (see #306) ... this small change definitely makes sense and was just an oversight

PR #570, opened 2023-07-25
feat: support Pydantic2 (as well as Pydantic1)

tlambert03:


jni:
@tlambert03 Just for casual observers it might be soothing to rename this PR to "Support both Pydantic v1 and v2" 😅

tlambert03:
surely you don't think the guy who tried to raise the alarm bells about the dangers of hard-pinned pydantic2 requirements would just move to pydantic2 in a package that napari uses! 😅

PR #571, opened 2023-07-26
Auto-generated examples gallery

GenevieveBuckley:
Currently there are a ton of great examples in the repo that are effectively invisible to anyone looking at the main documentation page at https://pyapp-kit.github.io/magicgui/

This PR uses [mkdocs-gallery](https://smarie.github.io/mkdocs-gallery/) to automatically generate an examples gallery for the docs website. This provides some nice features, including:
* Buttons for users to download code in python file or jupyter notebooks format
* The CI will build this along with the rest of the docs, using the same mkdocs commands as before
* The code is executed to produce screenshots, so if your examples code somehow falls out of sync with the library syntax, you should get an error and warning your about that.
* if somebody adds a new example script, it is also automatically added to the website - no more accidentally forgetting to add new pages to the mkdocs nav table of contents!

I've previously done something similar for the napari-threedee project https://github.com/napari-threedee/napari-threedee/issues/99. If you'd like to play around with that live on their website, it's here: https://napari-threedee.github.io/generated/gallery/

Here's a screenshot of the proposed new examples page for magicgui, running locally on my laptop. There are a lot of examples, so I've tried to group them into meaningful themes.

screenshot-magicgui-examples-gallery


tlambert03:
😍🤩

Thank you!!  Will have a closer look through your comments soon

tlambert03:
I've had a look (and pushed a couple changes, so pull again before working on it more).  I really love it, thanks so much.

The only change I'm not thrilled about are the napari examples, and basically anything where the `# %%` syntax had to be added to the python source files.  I find them rather messy to read through, and feel like they kinda ruin the example files themselves (in favor of only ever looking at them in the built website).  I know it's a tradeoff, but i'd like to think about those a bit, or perhaps fall back to a hybrid mixture of examples.

But, I won't think about it too long, this is awesome as is and i want to get it in :)

GenevieveBuckley:
Thank you for figuring out the CI and formatting!

Two more things I noticed:
1. The napari example auto-generated screenshots don't show the magicgui dock widget. I don't know why, taking a screenshot manually seems to work. Maybe this doesn't matter too much, your gif movies are much better demonstrations in any case.
2. The progress bar examples don't show the progress bar in the auto-generated screenshots either. This is more understandable, since it doesn't appear until you click the "Run" button, which obviously isn't happening in the auto-generated docs. Again, maybe not an actual problem, just wanted to mention it.

GenevieveBuckley:
> The only change I'm not thrilled about are the napari examples, and basically anything where the `# %%` syntax had to be added to the python source files. I find them rather messy to read through, and feel like they kinda ruin the example files themselves (in favor of only ever looking at them in the built website). I know it's a tradeoff, but i'd like to think about those a bit, or perhaps fall back to a hybrid mixture of examples.

That's a reasonable hesitation. I've fiddled with things a bit, and I think things are a bit nicer now. 

This affects four files:
1. values_dialog.py
2. The magicgui jupyter notebook example
3. napari example - image arithmetic
4. napari example - parameter sweep

1 and 2 are because it's hard to run those scripts in the automatic CI. 
3 and 4 are because you want to combine code with text eplanation (but don't like the jupytext formatting that leaves a lot of `#` symbols on all the text lines).

First I tried to see if we could just add markdown files side by side along with the python example files, but it looks like that is not supported. If that's potentially a big long term issue, we could [ask if there are any future plans to support that behaviour in mkdocs-gallery](https://github.com/smarie/mkdocs-gallery/issues/new).

For now, what I've done is to move as much content as possible into the docstrings at the top of the file. Any code we actually want to execute goes below that. That at least cuts down on the number of lines prefaced with `#` when you download and look at the plain text.

You do still have to have at least one `# %%` line at the very bottom, otherwise you get a big "broken" text image for that example in the gallery. But if an empty `# %%` is there at the bottom of the file, it is treated like an empty code cell, and everything is fine.

Still a small compromise, but perhaps one we can live with for now. I personally think it helps address those concerns a bit.

pyproject.toml
```diff
@@ -197,7 +201,9 @@ extend-ignore = [
 
 [tool.ruff.per-file-ignores]
 "tests/*.py" = ["D", "E501"]
-"examples/*.py" = ["D", "B"]
+"docs/examples/*.py" = ["D", "B"]
+"docs/examples/napari/*" = ["E501"]
```
GenevieveBuckley, line 205: Turns out the inline `noqa`statements were being rendered as text in the actual website. This seems like a good way to avoid that problem.

docs/examples/applications/values_dialog.py
```diff
@@ -0,0 +1,21 @@
...
+vals = request_values(
+    age=int,
+    name={"annotation": str, "label": "Enter your name:"},
+    title="Hi, who are you?",
+)
+print(repr(vals))
+```
+"""
+
+# %%
```
GenevieveBuckley, line 21: Here is an example of an almost, but not quite entirely useless `# %%`
If it is missing, mkdocs-gallery freaks out and puts a big "BROKEN" image in the examples gallery. If we have the `# %%` line here, it is treated like an empty code cell with no output, which it is. Then you get the default not-scary image next to this example in the gallery page.

tlambert03 (approved):
let's get this in.  I think the end result is great.

there are still a number of things i'm not crazy about, but can address them in follow up PRs.  some notes so I remember:
- the gallery plugin appears to be regenerating all of the examples every time any file changes, so using `mkdocs serve` for realtime editing of the docs becomes much less useful.
- the gallery plugin appears to be using threading somehow, and sending an interrupt signal with control-C no longer shuts down the `mkdocs serve` live server.  this is a big pain... and the only way I've been able to stop it is to kill the entire terminal process.  It makes me a bit concerned with the stability/innards of the plugin itself.  
- I do have general concerns about the mkdocs-gallery plugin itself. in addition to the odd mkodcs serve behavior, it has a mkdocs deprecation warning from something that changed way back in Jun, 2021, so i wonder how actively maintained it is... and it's not a "popular" plugin (by stars) so perhaps has more rough edges.  
- i do prefer the current napari examples (with the docs at the top)... but it comes at the cost of code duplication between the code in the docstrings and the code itself, which now has the risk of going out of sync...  Perhaps includes/snippets could be used here, but I'll look into that later.

PR #572, opened 2023-07-28
Replace QSpinBox with QLargeIntSpinBox

gselzer:
The range of QSpinBox is limited to a 32-bit signed integer - Python integers can go much larger than that. By using QLargeIntSpinBox instead, we enable larger range options!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ min-req = [
     "docstring_parser==0.7",
     "psygnal==0.5.0",
     "qtpy==1.7.0",
-    "superqt==0.4.0",
+    "superqt==0.5.0",
     "typing_extensions",
 ]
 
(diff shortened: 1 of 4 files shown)


tlambert03 (approved):
Thanks!  LGTM assuming everything passes

tlambert03:
ok, this is pretty much ready to go.  The failing napari/stardist/partseg tests are all because napari installs are failing at the moment, and unrelated to magicgui.  The ubuntu-latest (3.9) pyqt5 is a legitimate superqt issue, but since it only appears to (repeatedly) happen on the specific combination of ubuntu/pyqt5/python3.9, it's a bit hard to test and fix.  might just skip it

tlambert03 (approved):
thanks @gselzer!

PR #588, opened 2023-09-30
Support Annotated types in list/tuple

hanjinliu:
I found that the annotated type hints such as `list[Annotated[...]]` and `tuple[Annotated[...], ...]` were dropped during the process of type resolution, so it was a bit inconsistent to configure `ListEdit` and `TupleEdit`.
This PR will fix it.


tlambert03 (approved):
thanks @hanjinliu!

PR #592, opened 2023-10-05
MAINT: Add pint as a requirement

larsoner:
On my system `pip intstall magicgui` followed by running a variant of the "all of these" example from https://pyapp-kit.github.io/magicgui/widgets/#magicgui I got:
```
monolith:Desktop larsoner$ python -ui magic.py 
  File "/Applications/MNE-Python/1.5.0_1/.mne-python/lib/python3.11/site-packages/superqt/spinbox/_quantity.py", line 4, in <module>
    from pint import Quantity, Unit, UnitRegistry
ModuleNotFoundError: No module named 'pint'

The above exception was the direct cause of the following exception:

...
  File "/Applications/MNE-Python/1.5.0_1/.mne-python/lib/python3.11/site-packages/superqt/__init__.py", line 57, in __getattr__
    from .spinbox._quantity import QQuantity
  File "/Applications/MNE-Python/1.5.0_1/.mne-python/lib/python3.11/site-packages/superqt/spinbox/_quantity.py", line 7, in <module>
    raise ImportError(
ImportError: pint is required to use QQuantity.  Install it with `pip install pint`
```
I guess `pint` is an optional dependency for `superqt` (?) but it seems like `magicgui` should require it to make sure all examples work out of the box.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
         "psygnal>=0.5.0",
         "qtpy>=1.7.0",
         "superqt>=0.5.0",
+        "pint",
         "typing_extensions",
     ],
 )


tlambert03:
hey @larsoner :wave: nice to see you here  and thanks for opening a PR.

I'm hesitant to add more base level dependencies to magicgui.  By its nature, magicgui might potentially support _many_ third party types (beyond pint) but shouldn't require those packages for installation.  My assumption with pint was that, if someone wants to provide a widget representation of a `pint.Quantity`, then that user would be declaring a dependency on pint already.  We do add a `quantity` extra (`pip install magicgui[quantity]`) along with a number of other optional types:

https://github.com/pyapp-kit/magicgui/blob/21ee4524c9b0ffc8cd3253f6e3cc080008c89d37/pyproject.toml#L56-L66

I see that this means that someone can't immediately run all of the examples without additional installations, but if we add pint, then a similar argument could be made for `pillow`, `tqdm`, and any number of third party types that we're able to support.

As an alternative, would you support the addition and of an `all` or `examples` extra that brings in all of these optional dependencies for someone wanting to run all of the examples?

larsoner:
> would you support the addition and of an all or examples extra

Actually I didn't even notice those options. Maybe just a `.. note::` or `.. warning::` where the `QuantityEdit` shows up on that examples page and also the [QuantityEdit API page](https://pyapp-kit.github.io/magicgui/api/widgets/QuantityEdit/#magicgui.widgets.QuantityEdit) would help to know about when optional dependencies are needed. This would also be relevant for the `matplotlib` stuff we've been discussing.

Just let me know if you agree with how and where these should be documented (or if you have other ideas) and I'm happy to adjust this PR to be a documentation update instead 👍

tlambert03:
> Just let me know if you agree with how and where these should be documented (or if you have other ideas) and I'm happy to adjust this PR to be a documentation update instead

agree! 
A note could be added to the doc header of [this file](https://github.com/pyapp-kit/magicgui/blob/main/docs/examples/applications/pint_quantity.py#L7) and also to the docstring [here](https://github.com/pyapp-kit/magicgui/blob/main/src/magicgui/widgets/_concrete.py#L1041).

i just tried locally, and mkdocs admonition syntax (`!!note`) doesn't appear to be working in the gallery pages at the moment. so if you just make a simple note i can fix the admonitions later

tlambert03:
(note also: I just killed the antiquated setup.py file in #595 ... so your conflict is due to that.  it can just be removed)

PR #597, opened 2023-10-09
feat: add toolbars [wip]

### LINKED ISSUE #593: Add support for toolbars
It is possible to create toolbars on both Qt and in notebooks, e.g, this is what we have in MNE-Python:

It would be great if magicgui could support both. I might be up for implementing these since I already have some relevant abstraction code.

tlambert03:
closes #593 

this PR includes #598, and should wait for it to merge


tlambert03:
this is mostly working now for both backends... but doesn't yet have icon support.  I might want to add icons in a new PR.  superqt recently [got support](https://github.com/pyapp-kit/superqt/pull/209) for [iconify](https://iconify.design) ... and it would nice to be able to provide the same flexibility for ipywidgets.

tlambert03:
@larsoner, i know haven't dug too deeply into how things are structured here, but if you have a moment to look at the general [protocol](https://github.com/pyapp-kit/magicgui/pull/597/files#diff-a0d809a2eec29106f71c8874fd2d0937ba2e1bceb2a3923ebfdbcd5bc6440a8cR505-R538), and backend implementations here ([ipynb](https://github.com/pyapp-kit/magicgui/pull/597/files#diff-9aefa2a3555b653495ec52b1ae77b78433d0db03cf3a7e98deb1d50a91ee187b), [qt](https://github.com/pyapp-kit/magicgui/pull/597/files#diff-494814f6ad0b77deade96d5dec811e12a5e789c5641ad0d21e38059adff98472)), I'd welcome your input.

I implemented a toolbar as a top level widget (`magicgui.widgets.Toolbar()`) rather than a method on something like main window or layout.  So it would be used by adding/inserting a `Toolbar()` into a `widgets.Container`

still need to write tests, and as mentioned above, still contemplating how to add icons (see related https://github.com/pyapp-kit/magicgui/pull/598)

larsoner:
So far so good! To get a "native feel" I have generically:
```
toolbar = widgets.ToolBar()
noop = lambda *args, **kwargs: None
toolbar.add_button("Play", icon="play", callback=noop)
toolbar.add_button("Pause", icon="pause", callback=noop)
toolbar.add_button("Screenshot", icon="camera", callback=noop)
toolbar.add_button("Movie", icon="film", callback=noop)
```
then in my Qt half doing:
```
    window = widgets.MainWindow(widgets=columns, layout="horizontal", labels=False)
    window.native.addToolBar(toolbar.native)
```
and in the NB half doing:
```
    from ipywidgets import AppLayout
    window = AppLayout(
        header=toolbar.native,
        left_sidebar=columns[0].native,
        center=columns[1].native,
        pane_heights=['30px', 5, 1]
    )
```
plus the rest of the stuff from https://github.com/mne-tools/mne-python/issues/11990 I get a decent result (other than Qt icons as expected) except for I think the toolbar button text being wrong on Qt?:

| NB | Qt |
| -- | -- |
| Screenshot from 2023-10-10 13-48-51 | Screenshot from 2023-10-10 13-47-41 |

tlambert03:
thanks! 

> except for I think the toolbar button text being wrong on Qt?

 ah yep, that's related to icons being unfinished.  will fix

question for you regarding API:
I'm inclined to add an explicit `MainWindow.add_toolbar(tb: widgets.ToolBar)` method to take full advantage of `QMainWindow.addToolBar`.  Do you prefer that to a magic alternative where we call that under the hood in the qt backend if a ToolBar is in the list of the widgets? (or perhaps just the _first_ item in the list)?

larsoner:
I think the explicit one is better and what I would expect.

tlambert03:
i'm going to merge this and continue work in #604

PR #598, opened 2023-10-10
feat: add icons on buttons

tlambert03:
this PR allows adding icons to buttons, it uses superqt's recently added iconify support (https://github.com/pyapp-kit/superqt/pull/209) and ipywidget's built in support for fontawesome 5.  So you can do something like this:

```python
from magicgui import widgets

btn = widgets.PushButton(icon="bi:cloud")
btn.show(run=True)
```

Note that qt will have the full selection of [all icons available in iconify](https://icon-sets.iconify.design), but ipywidgets only works when the icon string after the colon is valid in fontawesome 4/5. In some cases (if you don't use font-awesome prefix as shown in the example above) that will mean a different appearance in qt.  If you want the icons to look similar in both backends, just restrict yourself to [fontawesome-5](https://icon-sets.iconify.design/fa-solid/)

| qt | ipwidgets|
|---|---------|
|Screen Shot 2023-10-10 at 1 18 29 PM| Screen Shot 2023-10-10 at 1 19 53 PM|

### DIFF
--- a/src/magicgui/backends/_ipynb/widgets.py
+++ b/src/magicgui/backends/_ipynb/widgets.py
@@ -260,11 +260,24 @@ class _IPySupportsText(protocols.SupportsText):
         return self._ipywidget.description
 
 
+class _IPySupportsIcon(protocols.SupportsIcon):
+    """Widget that can show an icon."""
+
+    _ipywidget: ipywdg.Widget
+
+    def _mgui_set_icon(self, value: str, color: str) -> None:
+        """Set icon."""
+        # not all ipywidget buttons support icons (like checkboxes),
+        # but our button protocol does.
+        if hasattr(self._ipywidget, "icon"):
+            self._ipywidget.icon = value.replace("fa-", "")
+
+
 class _IPyCategoricalWidget(_IPyValueWidget, _IPySupportsChoices):
     pass
 
 
-class _IPyButtonWidget(_IPyValueWidget, _IPySupportsText):
+class _IPyButtonWidget(_IPyValueWidget, _IPySupportsText, _IPySupportsIcon):
     pass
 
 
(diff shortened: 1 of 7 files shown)


src/magicgui/backends/_qtpy/widgets.py
```diff
@@ -433,12 +442,48 @@ def _mgui_get_text(self) -> str:
...
+    if ":" not in key:
+        # for parity with the other backends, assume fontawesome
+        # if no prefix is given.
+        key = f"fa:{key}"
+
+    try:
+        return superqt.QIconifyIcon(key, color=color)
+    except (OSError, ValueError) as e:
+        warnings.warn(f"Could not set iconify icon: {e}", stacklevel=2)
+        return None
```
Czaki, line 480: Does this will handle HTTP errors if someone try to use it the first time on a machine without internet access?

Czaki (commented):
it looks nice. 
Shor check of ipywidgets code suggests that it may be possible to easily contribute to enable arbitrary SVG on buttons.

src/magicgui/backends/_qtpy/widgets.py
tlambert03: Yeah I believe it should (that was my intention anyway).  Requests HTTP errors are subclasses of OSError

src/magicgui/backends/_ipynb/widgets.py
```diff
@@ -1,3 +1,5 @@
+# from __future__ import annotations  # NO
```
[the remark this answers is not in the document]
tlambert03: Oh just a silly reason, cause I use the annotations at runtime internally in this module.  I can add a comment

tlambert03:
> it looks nice. 
> 
> Shor check of ipywidgets code suggests that it may be possible to easily contribute to enable arbitrary SVG on buttons.

You mean like an upstream PR?  That's a good idea

tlambert03:
I opened https://github.com/jupyter-widgets/ipywidgets/issues/3850 for discussion, but probably won't wait here for it

PR #600, opened 2023-10-10
chore: preserve magicgui-decorated function parameter hints with ParamSpec

tlambert03:
This PR uses ParamSpec to preserve function type hints when decorated with `@magicgui`

before this PR:

Screen Shot 2023-10-10 at 7 24 23 PM

after:

Screen Shot 2023-10-10 at 7 24 06 PM

### DIFF
--- a/src/magicgui/type_map/_magicgui.py
+++ b/src/magicgui/type_map/_magicgui.py
@@ -16,17 +16,22 @@ from typing import (
 from magicgui.widgets import FunctionGui, MainFunctionGui
 
 if TYPE_CHECKING:
+    from typing_extensions import ParamSpec
+
     from magicgui.application import AppRef
+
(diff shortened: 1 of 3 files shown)


multimeric:
Hmm, I think this may have broken some downstream things. I believe adding an extra generic parameter to your classes changes the inheritance interface, because inheriting classes now have to add the corresponding parameter or it will be a runtime error: `TypeError: Too few arguments for <class '__main__.Foo'>; actual 1, expected 2`. Happy to be convinced otherwise.

The commit in question is: https://github.com/pyapp-kit/magicgui/commit/05c92921a5a6e7dad3757557e73d09c2f559ac01

tlambert03:
It would be helpful if you could include the code you ran with `Foo` in it so I can see exactly what you're trying to do/inherit.

PR #602, opened 2023-10-20
fix: Allow user overwritte default widget opts

Czaki:
This PR fixes https://github.com/napari/napari/issues/6354

It also renamems multiple variables starting from `_` to ending with `_` as it is convention to use leading underscore as an unused variable when it is a local variable in functions so make it harder for me to follow flow of code.

### DIFF
--- a/src/magicgui/type_map/_type_map.py
+++ b/src/magicgui/type_map/_type_map.py
@@ -214,8 +214,8 @@ def _pick_widget_type(
     raise_on_unknown: bool = True,
 ) -> WidgetTuple:
     """Pick the appropriate widget type for ``value`` with ``annotation``."""
-    annotation, _options = _split_annotated_type(annotation)
-    options = {**_options, **(options or {})}
+    annotation, options_ = _split_annotated_type(annotation)
+    options = {**options_, **(options or {})}
     choices = options.get("choices")
 
     if is_result and annotation is inspect.Parameter.empty:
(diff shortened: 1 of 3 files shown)


Czaki:
It looks like a failure of the docs is unrelated to PR.

tlambert03:
Since the pr is mostly rename now, can you point out which part specifically fixes the napari issue ?

src/magicgui/type_map/_type_map.py
```diff
@@ -269,14 +269,14 @@ def _pick_widget_type(
...
         return wdg, options
 
-    _widget_type = match_type(_type, value)
-    if _widget_type:
-        _cls, opts = _widget_type
-        return _cls, {**options, **opts}
+    widget_type_ = match_type(type_, value)
+    if widget_type_:
+        cls_, opts = widget_type_
+        return cls_, {**opts, **options}
```
Czaki, line 275: this

src/magicgui/type_map/_type_map.py
```diff
@@ -252,15 +252,15 @@ def _pick_widget_type(
...
 
     if is_result:
-        _widget_type = match_return_type(_type)
-        if _widget_type:
-            _cls, opts = _widget_type
-            return _cls, {**options, **opts}
+        widget_type_ = match_return_type(type_)
+        if widget_type_:
+            cls_, opts = widget_type_
+            return cls_, {**opts, **options}
```
Czaki, line 263: and here same change, but it is not the code branch reached by the original issue.

Czaki:
the change is to change `{**options, **opts}` to `{**opts, **options}`

tlambert03:
thanks!

PR #610, opened 2023-10-29
feat: add `Table.delete_row` method

tlambert03:
cc @jni 

would appreciate your input on the API.  I considered using `Table.drop` like the pandas API, but I thought it might be confusing that this one is in-place and that one is not.  I'm also wary of the confusion between row indices and row headers (both of which could conceivably be integers), so I forced the usage of a kwarg.

### DIFF
--- a/src/magicgui/widgets/_table.py
+++ b/src/magicgui/widgets/_table.py
@@ -269,6 +269,39 @@ class Table(ValueWidget, _ReadOnlyMixin, MutableMapping[TblKey, list]):
         for row, d in enumerate(data):
             self._set_rowi(row, d)
 
+    def delete_row(
+        self,
+        *,
+        index: int | Sequence[int] | None = None,
+        header: Any | Sequence[Any] | None = None,
+    ) -> None:
+        """Delete row(s) by index or header.
+
+        Parameters
+        ----------
+        index : int or Sequence[int], optional
+            Index or indices of row(s) to delete.
+        header : Any or Sequence[Any], optional
+            Header or headers of row(s) to delete.
+        """
+        indices: set[int] = set()
+        if index is not None:
+            if isinstance(index, Sequence):
+                indices.update(index)
+            else:
+                indices.add(index)
+        if header is not None:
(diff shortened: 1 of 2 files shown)


src/magicgui/widgets/_table.py
```diff
@@ -269,6 +269,39 @@ def value(self, value: TableData) -> None:
...
+            Header or headers of row(s) to delete.
+        """
+        indices: set[int] = set()
+        if index is not None:
+            if isinstance(index, Sequence):
+                indices.update(index)
+            else:
+                indices.add(index)
+        if header is not None:
+            if isinstance(header, str) or not isinstance(header, Sequence):
```
jni, line 294: Oh man, I needed a double-take to understand why this wasn't just `not isinstance(header, Sequence)`. That's annoying 😂

jni:
Missing a couple of test coverage lines, but looks good to me. I basically agree with your assessment. I wouldn't look to the pandas API but the draft [DataFrame API](https://data-apis.org/dataframe-api/draft/API_specification/index.html), which bizarrely has `drop_columns` but not `drop_rows` — perhaps they are leaking implementation details into their API choices? 😂 I also couldn't find an issue in their [repo](https://github.com/data-apis/dataframe-api). But anyway, `drop_columns` [appears to be in-place](https://data-apis.org/dataframe-api/draft/API_specification/dataframe_object.html#dataframe_api.DataFrame.drop_columns) (returns Self), so maybe `drop_rows` would be a good analogous method name? And maybe we should make an issue in their repo, though I don't feel super strongly about it.

src/magicgui/widgets/_table.py
tlambert03: Yep, classic :) string is a sequence but not "that" kind :)

tlambert03:
Thanks!

Btw: I would assume that it's in place because of `-> Self`.  That means it returns "type of self" (not actual self)

See this type of usage for example: https://peps.python.org/pep-0673/#use-in-classmethod-signatures

jni:
Fascinating! Thanks for the info!

PR #628, opened 2024-02-06
fix: fix parent_changed signal, and rename to native_parent_changed

### LINKED ISSUE #627: No parent with parent_changed
**Describe the bug**

The `parent` attribute seems to be always None and event handlers for `parent_changed` receive None.

**To Reproduce**

```python
import napari
viewer = napari.viewer.Viewer()
dock = Container()
def on_parent_changed(parent):
    print(f"parent_changed: {parent=}, {dock.parent=}")
    if parent is not None: # and visible etc.
        print("setup")
    else:
        print("cleanup")
dock.parent_changed.connect(on_parent_changed)
qt_dock = viewer.window.add_dock_widget(dock)
# Expected: "parent_changed: parent=<QtViewerDockWidget>, dock.parent=<QtViewerDockWidget>"
assert dock.parent is not None
# Click the close button, or remove the dock programmatically
viewer.window.remove_dock_widget(qt_dock)
# Expected: "parent_changed: parent=None, dock.parent=None"
assert dock.parent is None
```

**Expected behavior**

When `parent_changed` is triggered on hooking a widget into the Napari widgets tree, the parent should be passed and the parent attribute should be set. When the parent is removed, None should be passed.

Specifically, the example code would print
```
parent_changed: parent=<napari._qt.widgets.qt_viewer_dock_widget.QtViewerDockWidget object at 0x7fe02ce5a310>, dock.parent=<napari._qt.widgets.qt_viewer_dock_widget.QtViewerDockWidget object at 0x7fe02ce5a310>
setup
parent_changed: parent=None, dock.parent=None
cleanup
```
This works in napari<=0.4.19 and magicgui==0.7.3, but breaks with magicgui>=0.8.0.

**Context**

My use case is that my plugin's dock is supposed to do setup and cleanup when the dock becomes visible or is removed. This was previously necessary because dock instances were just hidden and reused when opening again. I had code that would listen for the _visibilityChanged_ event on the Qt parent, which I can only register after a parent was added. 

What I am really interested is being notified when the dock is shown or not shown anymore.
Possibly there is now a better way to do it, or a Napari API with dock activated/visible and deactivated/hidden hooks (I am not aware of).

 - OS: Ubuntu 22.04.3 LTS
 - backend: PyQt5==5.15.10
 - magicgui==0.8.1

tlambert03:
This fixes #627 by properly including the `widget.native.parent()` as the first argument of the signal.  

However, that signal name is rather confusing, since `Widget.parent` now mean a `magicgui.widgets.Container` (see #582 and #456).  So, this PR also renames it as `native_parent_changed`, to make clearer that what the callback will receive is the native object and *not* a magicgui container parent.  (`parent_changed` is retained as a deprecated alias). 

This means that after `widget.native_parent_changed` has been emitted with a non-null value, it is still quite possible that `widget.parent` will be `None` if it is not inside of a magicgui Container.  However, `widget.native.parent()` will not be None.

cc @aeisenbarth.  hope this will help you.  let me know if you have any thoughts


aeisenbarth:
Looks good to me, thank you!

PR #638, opened 2024-03-14
Add Separator to ComboBox

qin-yu:
Hey @tlambert03, 

I looked into `magicgui` and added a very simple mechanism for `ComboBox` to add separators by introducing a `Separator` class. Please have a look and tell me if you like it or not. Do I need to do the documentation somewhere or someone is in charge of this?

Many thanks,
Qin

### Example

By giving a list `["a", Separator(), "b"]` to magicgui for creating a dropdown list, the `ComboBox` backend go through the list, check if the data is a `Separator` and insert a separator there using `.insertSeparator()`.

I put a modified version of the example script from your repo:

```python
from magicgui import magicgui
from magicgui.types import Separator

sep = [Separator(i) for i in range(4)]
a = [1, 2, sep[0], 4, sep[1], 6, 7, sep[2], 9, sep[3], 11, 12, sep[0], 14, sep[0]]
b = [1, 2, sep[0], 4, sep[1], 6, 7, sep[2], 9, sep[3], 6, 7, sep[0], 9, sep[0]]


@magicgui(call_button="Test Separator",
          a={"widget_type": "ComboBox", "choices": a},
          b={"widget_type": "ComboBox", "choices": b},
          result_widget=True)
def test_separator(a=a[0], b=b[0]):
    return


test_separator.show(run=True)
```

`a` with unique elements:

image

`b` with duplicates:

image

### DIFF
--- a/src/magicgui/backends/_ipynb/widgets.py
+++ b/src/magicgui/backends/_ipynb/widgets.py
@@ -221,7 +221,6 @@ class _IPySupportsChoices(protocols.SupportsChoices):
     def _mgui_set_choices(self, choices: Iterable[tuple[str, Any]]) -> None:
         """Set available choices."""
         options = [item for item in choices if item[1] is not Separator]
-        options = list(dict(options).items())  # remove duplicate names, last data used
         self._ipywidget.options = options
 
     def _mgui_del_choice(self, choice_name: str) -> None:
--- a/src/magicgui/backends/_ipynb/widgets.py
+++ b/src/magicgui/backends/_ipynb/widgets.py
@@ -12,6 +12,7 @@ except ImportError as e:
     ) from e
 
 
+from magicgui.types import Separator
 from magicgui.widgets import protocols
 
 if TYPE_CHECKING:
@@ -219,7 +220,9 @@ class _IPySupportsChoices(protocols.SupportsChoices):
 
     def _mgui_set_choices(self, choices: Iterable[tuple[str, Any]]) -> None:
         """Set available choices."""
-        self._ipywidget.options = choices
+        options = [item for item in choices if item[1] is not Separator]
+        options = list(dict(options).items())  # remove duplicate names, last data used
+        self._ipywidget.options = options
 
     def _mgui_del_choice(self, choice_name: str) -> None:
         """Delete a choice."""
@@ -248,7 +251,7 @@ class _IPySupportsChoices(protocols.SupportsChoices):
 
     def _mgui_set_choice(self, choice_name: str, data: Any) -> None:
         """Set the data associated with a choice."""
-        self._ipywidget.options = (*self._ipywidget.options, (choice_name, data))
+        pass
 
 
 class _IPySupportsText(protocols.SupportsText):
--- a/src/magicgui/backends/_ipynb/widgets.py
+++ b/src/magicgui/backends/_ipynb/widgets.py
@@ -250,7 +250,7 @@ class _IPySupportsChoices(protocols.SupportsChoices):
 
     def _mgui_set_choice(self, choice_name: str, data: Any) -> None:
         """Set the data associated with a choice."""
-        pass
+        self._ipywidget.options = (*self._ipywidget.options, (choice_name, data))
 
 
 class _IPySupportsText(protocols.SupportsText):
--- a/src/magicgui/backends/_qtpy/widgets.py
+++ b/src/magicgui/backends/_qtpy/widgets.py
@@ -908,7 +908,7 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
         self._event_filter.valueChanged.connect(callback)
 
     def _mgui_get_count(self) -> int:
-        """Return the number of items in the dropdown."""
+        """Return the number of items in the dropdown, including separator items."""
         return self._qwidget.count()
 
     def _mgui_get_choice(self, choice_name: str) -> Any:
@@ -930,13 +930,18 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
 
     def _mgui_set_choice(self, choice_name: str, data: Any) -> None:
         """Set data for ``choice_name``."""
-        item_index = self._qwidget.findText(choice_name)
-        # if it's not in the list, add a new item
-        if item_index == -1:
-            self._qwidget.addItem(choice_name, data)
-        # otherwise update its data
+        if data is Separator:
+            item_index = self._mgui_get_count()
+            self._qwidget.insertSeparator(item_index)  # itemData is None
+            self._qwidget.setItemData(item_index, Separator)
         else:
-            self._qwidget.setItemData(item_index, data)
+            item_index = self._qwidget.findText(choice_name)
+            # if it's not in the list, add a new item
+            if item_index == -1:
+                self._qwidget.addItem(choice_name, data)
+            # otherwise update its data
+            else:
+                self._qwidget.setItemData(item_index, data)
 
     def _mgui_set_choices(self, choices: Iterable[tuple[str, Any]]) -> None:
         """Set current items in categorical type ``widget`` to ``choices``."""
@@ -954,11 +959,7 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
                     self._qwidget.removeItem(i)
             # update choices and insert separators
             for name, data in choices_:
-                if data is Separator:
-                    self._qwidget.insertSeparator(self._mgui_get_count())
-                else:
-                    self._mgui_set_choice(name, data)
-
+                self._mgui_set_choice(name, data)
             # if the currently selected item is not in the new set,
             # remove it and select the first item in the list
             current2 = self._qwidget.itemText(self._qwidget.currentIndex())
@@ -983,6 +984,7 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
         return tuple(
             (self._qwidget.itemText(i), self._qwidget.itemData(i))
             for i in range(self._qwidget.count())
+            if self._qwidget.itemData(i) != Separator
         )
 
 
--- a/src/magicgui/backends/_qtpy/widgets.py
+++ b/src/magicgui/backends/_qtpy/widgets.py
@@ -954,9 +954,8 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
                     self._qwidget.removeItem(i)
             # update choices and insert separators
             for name, data in choices_:
-                if isinstance(data, Separator):
-                    for _ in range(data.thickness):
-                        self._qwidget.insertSeparator(self._mgui_get_count())
+                if data is Separator:
+                    self._qwidget.insertSeparator(self._mgui_get_count())
                 else:
                     self._mgui_set_choice(name, data)
 
--- a/src/magicgui/backends/_qtpy/widgets.py
+++ b/src/magicgui/backends/_qtpy/widgets.py
@@ -26,7 +26,7 @@ from qtpy.QtGui import (
     QTextDocument,
 )
 
-from magicgui.types import FileDialogMode
+from magicgui.types import FileDialogMode, Separator
 from magicgui.widgets import Widget, protocols
 from magicgui.widgets._concrete import _LabeledWidget
 
@@ -952,9 +952,13 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
             for i in reversed(range(self._qwidget.count())):
                 if self._qwidget.itemText(i) not in choice_names:
                     self._qwidget.removeItem(i)
-            # update choices
+            # update choices and insert separators
             for name, data in choices_:
-                self._mgui_set_choice(name, data)
+                if isinstance(data, Separator):
+                    for _ in range(data.thickness):
+                        self._qwidget.insertSeparator(self._mgui_get_count())
+                else:
+                    self._mgui_set_choice(name, data)
 
             # if the currently selected item is not in the new set,
             # remove it and select the first item in the list
--- a/src/magicgui/backends/_qtpy/widgets.py
+++ b/src/magicgui/backends/_qtpy/widgets.py
@@ -908,8 +908,12 @@ class ComboBox(QBaseValueWidget, protocols.CategoricalWidgetProtocol):
         self._event_filter.valueChanged.connect(callback)
 
     def _mgui_get_count(self) -> int:
-        """Return the number of items in the dropdown, including separator items."""
-        return self._qwidget.count()
+        """Return the number of items in the dropdown, omitting any separator items."""
+        return sum(
+            1
+            for i in range(self._qwidget.count())
+            if self._qwidget.itemData(i) != Separator
+        )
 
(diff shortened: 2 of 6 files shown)


qin-yu:
Hey @tlambert03, I've completed the necessary changes, and I believe this Pull Request is now ready for your review. Tbh, this is my first time using Codecov, so I would greatly appreciate your feedback on its integration and any adjustments you might recommend. Thank you for your support!

tlambert03 (approved):
this looks excellent!  nice and simple too.  Thanks @qin-yu!

I think the only thing that gave me a little pause was the `thickness` parameter.  Since it's not really "thickness" per-se, but rather "number of".  Someone could also do `Sep(), Sep()` if they really wanted multiples, and Qt itself doesn't really have thickness. 

Did you add it cause it's something you know you want though?

qin-yu:
> Since it's not really "thickness" per-se, but rather "number of".  Someone could also do `Sep(), Sep()` if they really wanted multiples, and Qt itself doesn't really have thickness. 
> 
> Did you add it cause it's something you know you want though?   

Oops, I guess I went a bit too custom with my solution. I was working on this app using Napari, and I noticed that a single separator just wasn’t cutting it in terms of visibility. It took me a while to figure out that changing thickness/padding wasn’t straightforward. So, I ended up with this workaround in my own project. 

You’re right about using Sep(), Sep() for multiple separators. If you don't like fake thickness implementation, let's remove it and mention this workaround in docstrings.

Thanks for the heads-up! @tlambert03

tlambert03:
yeah, i'm not *totally* opposed to it, but maybe would prefer to consider alternatives before adding it to the public API :) let's just go with a parameter-free sentinel for now.

qin-yu:
Absolutely, I’m with you on maintaining a clean public API. I’ll go ahead and remove the parameter, then submit a new commit a bit later.

Maybe our conversation here could serve as a good reference for future users. Do you think it would be beneficial to include a note in the docstring?

tlambert03:
> Do you think it would be beneficial to include a note in the docstring?

sure, if you'd like, you could indicate that it can be used multiple times to create multiple lines 👍

tlambert03:
>  I was working on this app using Napari, and I noticed that a single separator just wasn’t cutting it in terms of visibility. It took me a while to figure out that changing thickness/padding wasn’t straightforward.

by the way, I see this as a problem with napari's style sheets, and something you could take up with napari developers.  And not something we should make special considerations for here in magicgui.

The default styling of separators actually looks quite nice, and in-line with the OS standards:

Screenshot 2024-03-17 at 3 23 04 PM

napari applies custom styling to everything, which makes it hard to see:

Screenshot 2024-03-17 at 3 22 33 PM

qin-yu:
> by the way, I see this as a problem with napari's style sheets, and something you could take up with napari developers. And not something we should make special considerations for here in magicgui.

Indeed, my experience with magicgui has been primarily within Napari, so my thoughts were anchored in that context 😢 

I've updated the code again. Given that the Separator sentinel would remain identical, I've made it a singleton too. I've also adjusted the tests accordingly and updated the docs. I removed a comma to keep the docstring <= 88 limit. Maybe it's the best thing we can do, but it looks okay.

image

src/magicgui/widgets/_concrete.py
```diff
@@ -342,7 +342,7 @@ def base(self, base: float) -> None:
 
 @backend_widget
 class ComboBox(CategoricalWidget):
-    """A dropdown menu, allowing selection between multiple choices."""
+    """A dropdown menu allowing ``Separator`` and selection between multiple choices."""
```
tlambert03, line 345:
```suggestion
    """A dropdown menu, allowing selection between multiple choices."""
```

i'm not sure it warrants first mention in the ComboBox description :)

tests/test_widgets.py
```diff
@@ -1059,6 +1059,38 @@ def f(x: Set[Lit]): ...
...
+        b={"widget_type": "ComboBox", "choices": b},
+        result_widget=True,
+    )
+    def widget_with_separator(a=a[0], b=b[0]):
+        return
+
+    assert widget_with_separator.a.native.count() == 21
+    assert widget_with_separator.b.native.count() == 18
+    assert widget_with_separator.a.choices == a2
+    assert widget_with_separator.b.choices == b2
```
tlambert03, line 1091: sorry, I missed this the first time.  I think this is a bit unfortunate actually.  Since the separator is a purely UI element (i.e. it can't actually be selected), I'm not sure it should be returned in the choices.  I think I would expect

```python
assert widget_with_separator.a.choices == (1, 2, 4, 6, 7, 9, 11, 12, 14)
```

thoughts?

tlambert03:
one more question:  can you perhaps look into how this could be implemented on the jupyter widgets backend?  by modifying `magicgui.backends._ipynb.widgets.ComboBox._mgui_set_choices`

src/magicgui/backends/_qtpy/widgets.py
```diff
@@ -952,9 +952,12 @@ def _mgui_set_choices(self, choices: Iterable[tuple[str, Any]]) -> None:
             for i in reversed(range(self._qwidget.count())):
                 if self._qwidget.itemText(i) not in choice_names:
                     self._qwidget.removeItem(i)
-            # update choices
+            # update choices and insert separators
             for name, data in choices_:
-                self._mgui_set_choice(name, data)
+                if data is Separator:
+                    self._qwidget.insertSeparator(self._mgui_get_count())
```
tlambert03, line 958: also, this conditional should probably move one level lower to `_mgui_set_choice`

tests/test_widgets.py
tlambert03: one way to accomplish this would be to `.setItemData(i, '<some-special-key>')` immediately after calling `insertSeparator`, and then filtering out those things in _mgui_get_choices

qin-yu: Absolutely, separators are items in Qt/QComboBox, so we can't avoid having them, and your `.setItemData(i, '<some-special-key>')` is very clever. You are also right about not returning all the `None`s 🤦🏻‍♂️I'll update `_mgui_get_choices()` and `_mgui_get_count()`.

qin-yu:
Thanks for the thorough reviews. I've just updated the `_qtpy` backend based on your feedback. Planning to dive into `magicgui.backends._ipynb.widgets.ComboBox._mgui_set_choices` first thing tomorrow. Really appreciate your guidance here!

qin-yu:
> one more question: can you perhaps look into how this could be implemented on the jupyter widgets backend? by modifying `magicgui.backends._ipynb.widgets.ComboBox._mgui_set_choices`

Regarding ipynb support nuances:
- I noticed ipynb allows for duplicate choice names with distinct data. Should we align this with the Qt backend's behavior to ensure consistency across platforms?
- Given Jupyter Widgets lacks native support for separators, I propose two approaches:
  1. Exclude separators entirely in `_mgui_set_choice()` to adhere to our principle of not incorporating features unsupported by the backend.
  2. Omit separators only within `_mgui_set_choices()` to parallel the Qt backend behavior. This way, a Separator in Qt equates to a conceptual Separator in ipynb, even though it's not visually represented.

What do you think? @tlambert03

qin-yu:
I've adjusted `magicgui.backends._ipynb.widgets.ComboBox`'s `_mgui_set_choices()` method to skip over Separators, given that Jupyter Widgets doesn’t natively support such a concept. I also fixed #639 in the same place, I hope the changes align with your design.

Regarding `_mgui_set_choice()` (without the ‘s’), it appears this method has no effects, so I've temporarily set it to `pass`. However, should there be a hidden use case or future need, here's a draft implementation I propose:

```python
def _mgui_set_choice(self, choice_name: str, data: Any) -> None:
    """Set the data associated with a choice."""
    if data is Separator:
        self._ipywidget.options = (*self._ipywidget.options, ("----", Separator))
    else:
        options = self._ipywidget.options
        for item in options:
            print(item, choice_name, data)
            if (isinstance(item, tuple) and item[0] == choice_name) or item == choice_name:
                options = [
                    item for item in options
                    if (not isinstance(item, tuple) or item[0] != choice_name) and item != choice_name
                ]
        self._ipywidget.options = (*options, (choice_name, data))
```

qin-yu:
(There is one failed test but I don't think it relates to this PR)

qin-yu:
image

qin-yu:
Hey @tlambert03 I encountered an error in the CI workflow [Test (3.10, ubuntu-latest, pyside6) / ubuntu-latest py3.10 pyside6](https://github.com/pyapp-kit/magicgui/actions/runs/8332884901/job/22803082860?pr=638#logs), and I'm not sure how to resolve it. Could you take a look and provide some guidance?

qin-yu:
All checks passed. It looks like we're in a good position to wrap up the PR. I'm assuming there's nothing further I need to do on my end if you're planning to handle the squash merge?

tlambert03:
thank you @qin-yu!  busy couple days here.  yes I will merge soon after having one more look.  thanks again!

tests/test_widgets.py
```diff
@@ -1059,6 +1059,59 @@ def f(x: Set[Lit]): ...
...
+    b2 = [1, 2, *sep[0], 4, *sep[1], 6, 7, *sep[2], 9, *sep[3], *sep[0], *sep[0]]
+
+    @magicgui(
+        call_button="Test Separator",
+        a={"widget_type": "ComboBox", "choices": a},
+        b={"widget_type": "ComboBox", "choices": b},
+        result_widget=True,
+    )
+    def widget_with_separator(a=a[0], b=b[0]):
+        return
```
tlambert03, line 1087: no need to bring `@magicgui` into this :) let's just make comboboxes:

```python
    combo_a = widgets.ComboBox(choices=a, value=a[0])
    combo_b = widgets.ComboBox(choices=b, value=b[0])
```

and use them below `widget_with_separator.a` -> `combo_a`

src/magicgui/backends/_qtpy/widgets.py
```diff
@@ -908,7 +908,7 @@ def _mgui_bind_change_callback(self, callback):
         self._event_filter.valueChanged.connect(callback)
 
     def _mgui_get_count(self) -> int:
-        """Return the number of items in the dropdown."""
+        """Return the number of items in the dropdown, including separator items."""
```
tlambert03, line 911: this actually becomes a bit problematic.  The protocol specifies that `_mgui_get_count` should return the number of _choices_:

https://github.com/pyapp-kit/magicgui/blob/781a54a3d0fb49845ededf734dbf30f71e18509c/src/magicgui/widgets/protocols.py#L412-L415

so, it's not really up to the backend here to change those semantics.   The `ComboBox` itself (via `CategoricalWidget`) declares `__len__` to be the number of choices (not the number of widgets)

https://github.com/pyapp-kit/magicgui/blob/781a54a3d0fb49845ededf734dbf30f71e18509c/src/magicgui/widgets/bases/_categorical_widget.py#L105-L107

so, `assert len(combo_a) == len(combo_a.choices)` should technically be true, but is not currently so in the tests.   So, I'm afraid this needs to account for that as well.  

I know these are annoying things, but this feature is a revealing itself to be a bit tricky to implement in a way that isn't potentially surprising and too magical.  It is blurring the lines between model (i.e. the values in `choices`) and the view (how they look in the combobox) a bit. 

i'm honestly not sure what the best thing to do here is.  one alternative is to go back to what you started with, and just allow for `choices` to include the separator.  that is, when you call `widget.choices` you *do* see all the `Separator` singletons in there (they would need to be subbed back in for `None`).   Or, we just fix the count issue as well and stick with it as it is.  That would then mean that it's basically impossible to query where the separators are, and whether they're in there (but you could also set them).  Which do you prefer?

qin-yu:
> i'm honestly not sure what the best thing to do here is. one alternative is to go back to what you started with, and just allow for choices to include the separator. that is, when you call widget.choices you do see all the Separator singletons in there (they would need to be subbed back in for None). Or, we just fix the count issue as well and stick with it as it is. That would then mean that it's basically impossible to query where the separators are, and whether they're in there (but you could also set them). Which do you prefer?

Indeed, the decision should be guided by the principle of least astonishment: opting for the approach that aligns with most developers' expectations and simplifies common tasks in ComboBox usage.

Although integrating separators as selectable items offers transparency and consistency—a notion shaped by my experiences with Qt, where separators are considered part of the item set—it's essential to acknowledge that separators are typically not viewed as actionable components. Therefore, omitting separators from counts and choices presents a more user-friendly approach, simplifying development by removing the need to distinguish between separators and actual items.

If you know that many of our users choose magicgui to bypass complexities in Qt, opting to conceal separators within the ComboBox would enhance simplicity. This design might require introducing new methods for managing item selection and modification. However, as a user, I would prefer simply updating the entire list (`choice=new_list`).

qin-yu:
Sorry. I said this may introduce new methods for item selection because I assumed it's selected by index.

`CategoricalWidget`s select items by name, allowing Qt duplicates in the future could be problematic, but I'll just simply hide the separators as the rest of concerns are out of the scope of this issue.

qin-yu:
It's pretty straightforward, actually! I just made the necessary adjustments to align with the protocol.

src/magicgui/backends/_ipynb/widgets.py
```diff
@@ -248,7 +250,7 @@ def _mgui_get_current_choice(self) -> str:
 
     def _mgui_set_choice(self, choice_name: str, data: Any) -> None:
         """Set the data associated with a choice."""
-        self._ipywidget.options = (*self._ipywidget.options, (choice_name, data))
+        pass
```
tlambert03, line 253: this change looks suspicious.  why are we removing the ability to set a single choice?  Won't this change mess up calling `CategoricalWidget.set_choice`?

tlambert03 (commented):
thanks again @qin-yu .  I think i'm happy to go with this now.  Please just address the remaining comment on the new change to `_IPySupportsChoices._mgui_set_choice` which i think needs to be reverted

src/magicgui/backends/_ipynb/widgets.py
qin-yu: You're absolutely right. I may have gotten hung up on the protocol's definition of choice_name update. While my commit seemed to function without breaking tests, this change was completely unnecessary. I'm changing it later.

qin-yu:
Hi @tlambert03, just pushed the final change for this PR! Apologies for the delay – I was recently on holiday. It's been a valuable learning experience collaborating with you on this PR. Working on a well-maintained Python package with comprehensive CIs like this is a great way to improve contribution practices.  Thanks for your patience and guidance!

tlambert03 (approved):
it was very pleasant working with you too!  Thanks a lot for the PR.  hope you'll do it again :)

qin-yu:
Hey Talley, please don't forget to make available v0.8.3 on conda-forge!

tlambert03:
looks like the release failed: https://github.com/pyapp-kit/magicgui/actions/runs/8476765032 ... sorry about that, will do again

qin-yu:
Thanks! But I think it failed again (actually no test can pass in the last few weeks)

tlambert03:
Yep I saw thanks,  I'll fix when I get the chance.  Won't be this weekend

qin-yu:
No worries, thanks for the information. Enjoy the rest of your weekend!

PR #640, opened 2024-03-23
feat: Better ListEdit UI

### LINKED ISSUE #636: ListEdit widget buttons should be nested and always use horizontal layout
When using a ListEdit for multiple selections of a dropdown, horizontal layout takes a *ton* of space, but changing it to vertical layout results in rather oversized buttons:

Screenshot 2024-03-14 at 6 10 02 pm

I think a sensible default here would be a vertical layout with the buttons nested inside a horizontal layout, but possibly other suggestions would also work.

hanjinliu:
Closes #636 

Vertical layout:
image

Horizontal layout:
image

As this PR changes the internal structure of ListEdit, people directly accessing the child widgets of ListEdit may be affected.

### DIFF
--- a/src/magicgui/widgets/_concrete.py
+++ b/src/magicgui/widgets/_concrete.py
@@ -6,7 +6,6 @@ super().__init__ calls.
 
 from __future__ import annotations
 
-import contextlib
 import datetime
 import inspect
 import math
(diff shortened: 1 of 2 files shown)


tlambert03:
looks great!  Thanks @hanjinliu

PR #651, opened 2024-06-05
Use PyPI trusted publishing to push release to PyPI

jni:


tlambert03 (commented):
Thanks @jni!  This will also need the id-token write permission added. See example in action docs

jni:
Oops! 😅 Well-spotted.

On napari we also have `contents: write` but that seems unnecessary?

tlambert03:
> Oops! 😅 Well-spotted.
> 
> 
> 
> On napari we also have `contents: write` but that seems unnecessary?

Thats necessary if you write releases and/or release notes using an action like softprops/action-gh-release

PR #652, opened 2024-06-14
Fix array data not compatible with ComboBox

hanjinliu:
After ComboBox supported separators, it does not allow `ndarray` and `DataFrame` as the choices anymore because of the comparison `!= Separator` (raises "ValueError: The truth value of a DataFrame is ambiguous."). 
This PR fixes it.


tlambert03 (approved):
thank you @hanjinliu!

PR #658, opened 2024-07-27
feat: Add mechanism to check basic subclassing of generics

Czaki:
This PR is addressing problem pointed by @psobolewskiPhD on napari zulip https://napari.zulipchat.com/#narrow/stream/309872-plugins/topic/List.20of.20LayerDataTuples.20yielded.20by.20a.20plugin.20not.20working.3F/near/454038830

The core of the problem is that if one registers `return` callback for `list[int]` it will not work with functions annotated with `typing.List[int]` and vice versa.  

This Pr is enhancing `safe_issubclass` to perform inheritance checking on generic as well. My goal is to allow use `Sequence[LayerDataTuple]` when registering type for magicgui widget and allow a user to specify `list[LayerDataTuple]`, `typing.List[LayerDataTuple]`, `tuple[LayerDataTuple, ...]` etc. 

I'm not sure if the impact of this solution on performance is significant. 

What is not suported. 

1) NamedTuple, when registered type, is plain `Tuple[something]`. 
2) Unions in `[]`

### DIFF
--- a/src/magicgui/type_map/_type_map.py
+++ b/src/magicgui/type_map/_type_map.py
@@ -104,10 +104,6 @@ def match_type(type_: Any, default: Any | None = None) -> WidgetTuple | None:
         args = get_args(type_)
         if len(args) == 1 and safe_issubclass(args[0], pathlib.Path):
             return widgets.FileEdit, {"mode": "rm"}
-        elif safe_issubclass(origin, list):
-            return widgets.ListEdit, {}
-        elif safe_issubclass(origin, tuple):
-            return widgets.TupleEdit, {}
 
     if safe_issubclass(origin, Set):
         for arg in get_args(type_):
--- a/src/magicgui/type_map/_type_map.py
+++ b/src/magicgui/type_map/_type_map.py
@@ -72,6 +72,7 @@ _SIMPLE_TYPES = {
     range: widgets.RangeEdit,
     slice: widgets.SliceEdit,
     Sequence[pathlib.Path]: widgets.FileEdit,
+    tuple: widgets.TupleEdit,
     Sequence: widgets.ListEdit,
     os.PathLike: widgets.FileEdit,
 }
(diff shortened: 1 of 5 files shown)


tlambert03:
seems reasonable to me, thanks @Czaki.  @hanjinliu, it appears that this PR might break the magic-class tests.  Could you look at the error and let me know if you think due to over-assumptions made on the magic-class side (which we should just ignore here), or whether it's indicative of something that could be a more general breaking change?

src/magicgui/type_map/_type_map.py
```diff
@@ -71,11 +71,14 @@ class MissingWidget(RuntimeError):
     datetime.datetime: widgets.DateTimeEdit,
     range: widgets.RangeEdit,
     slice: widgets.SliceEdit,
-    list: widgets.ListEdit,
-    tuple: widgets.TupleEdit,
+    Sequence[pathlib.Path]: widgets.FileEdit,
+    Sequence: widgets.ListEdit,
```
Czaki, line 75: maybe even more generic and use `Iterable` instead of `Sequence`?

Czaki:
> seems reasonable to me, thanks @Czaki. @hanjinliu, it appears that this PR might break the magic-class tests. Could you look at the error and let me know if you think due to over-assumptions made on the magic-class side (which we should just ignore here), or whether it's indicative of something that could be a more general breaking change?

It was a bug with handling sequence of Paths.

hanjinliu:
@Czaki , thank you for fixing this. At the time I install this branch it was already fixed 😅 
I accidentally found another bug, which I did not see before.

``` python
from magicgui import magicgui

@magicgui(x={"bind": [1]})
def func(x: list[int]):
    print(x)

func.show(True)
```

```
TypeError: ListEdit got an unexpected keyword argument: bind
```

I hit same error when I use `tuple[int, str]`. Do you have any idea why this is happening?

Czaki:
> @Czaki , thank you for fixing this. At the time I install this branch it was already fixed 😅 I accidentally found another bug, which I did not see before.

I see this error also on main. So it looks like some previous PR introduced it. 

I do not have experience with bind. Maybe @tlambert03 will know.

tlambert03:
yeah lets copy that new a new issue.  Thanks for digging into the sequence of paths thing @Czaki, I'll give this a final review later today then merge and can cut a new release if you'd like

tlambert03:
> I hit same error when I use tuple[int, str]. Do you have any idea why this is happening?

the quick answer is that `bind` is a parameter in ValueWidget, but list and tuple are container widgets (and do not inherit ValueWidget).

tlambert03 (approved):
LGTM

Czaki:
If you could inform on pointed zulip topic, when you release a new version?

PR #649, opened 2024-05-27
fix: fix compatibility with typing-extensions 0.4.12

tlambert03:
https://github.com/python/typing_extensions/pull/392 seems to be resulting in test errors:

```python
>       @magicgui

tests/test_widgets.py:932: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/type_map/_magicgui.py:190: in magicgui
    return _magicgui(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/type_map/_magicgui.py:530: in _magicgui
    return inner_func if function is None else inner_func(function)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/type_map/_magicgui.py:528: in inner_func
    return cast(FunctionGui, magic_class(func, **kwargs))
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/widgets/_function_gui.py:172: in __init__
    sig = magic_signature(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:319: in magic_signature
    return MagicSignature.from_signature(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:224: in from_signature
    return cls(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:204: in __init__
    params = [
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:205: in <listcomp>
    MagicParameter.from_parameter(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:169: in from_parameter
    return cls(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:109: in __init__
    _annotation = make_annotated(annotation, gui_options)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/magicgui/signature.py:72: in make_annotated
    return Annotated[annotation, _options]
[...]

PR #659, opened 2024-07-28
chore: Upload pydantic1 coverage with all other coverages

Czaki:
When I was working on #658 I spot that first message from codecov shows big drop of coverage. I was confused until I found that it is triggered by upload coverage of pydantic1 steep in separate. 

This PR unifies coverage upload.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -e .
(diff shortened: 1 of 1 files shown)


Czaki:
It works correctly and PR has single upload 
Screenshot 2024-07-28 at 11 30 52

tlambert03:
thanks @Czaki, i did this a bit differently in #660, using the existing pyapp-kit workflow approach (just forgot to include the pydantic1 step when I updated things in #649

Issue #661, opened 2024-07-30
"bind" incompatible with Container-like ValueWidgets


hanjinliu:
**Describe the bug**

Cannot use "bind" option with `ListEdit` and `TupleEdit` in the main branch version.
See https://github.com/pyapp-kit/magicgui/pull/658#issuecomment-2255844568

**To Reproduce**

``` python
from magicgui import magicgui

@magicgui(x={"bind": [1]})
def func(x: list[int]):
    print(x)

func.show(True)
```

```
TypeError: ListEdit got an unexpected keyword argument: bind
```

**Expected behavior**
"bind" should be processed similarly as other value widgets.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Windows
 - backend: PyQt5 5.15.10
 - magicgui version 0.8.4.dev12+g1479122

tlambert03:
The proper fix here is to create some sort of value widget container widget subclass.  I think the list widget and tuple widget were the first (only?) widgets that used Container but also implemented a duck-type ValueWidget interface (but without actually subclassing).  So if you want bind on those widgets, we should go back and create a proper value/container merge

hanjinliu:
`RangeEdit`, `SliceEdit` and `FileEdit` also inherits `Container`. I agree that these widgets should subclass `ValueWidget` at some points. In that sense, this issue will be a part of #372 .
I tried multiple inheritance and failed before, so I think we need a small refactoring of the widget classes. I will send a PR in a few day 😃

PR #663, opened 2024-08-01
Make all the valued containers subclass `ValueWidget`

### LINKED ISSUE #372: All value-like Container widgets should inherit from ValueWidget

hanjinliu:
Closes #372 #661 

Now this works:

``` python
from magicgui.widgets import FileEdit, PushButton
from magicgui.widgets.bases import ValueWidget

assert isinstance(PushButton(), ValueWidget)
assert isinstance(FileEdit(), ValueWidget)
```

I had to refactor more than I first thought. The major problem was that `ValueWidget` had `ValueWidgetProtocol` before, which is not compatible with container-like widgets (e.g. there's no `_mgui_get_value` for `FileEdit`). In this PR, I made `ValueWidget` more abstract: it only implements the basic interface for getting/setting values, thus independent of backends. `PrimitiveValueWidget` is the class that require backends, which is identical to the former `ValueWidget`. Besides, I introduced `_BaseContainerWidget` for `ValuedContainerWidget` because many methods of `ContainerWidget` is not publicly needed for value widgets (e.g. we never use `insert` for `FileEdit`).

Inheritance map is now like below.

```
Widget ----------------> ValueWidget --> PrimitiveValueWidget --> Label, LineEdit, Image etc.
   |                        |
   v                        v
_BaseContainerWidget --> ValuedContainerWidget -----------------> FileEdit, RangeEdit, ListEdit etc.
   |
   v
ContainerWidget ------------------------------------------------> Container
```

**Notes:**
- `EmptyWidget` is now a `ValuedContainerWidget ` because it's simply an empty container. Backend implementation of `EmptyWidget` is no longer needed.
- `ValueWidget` now inherits `ABC` because abstractmethod `get_value` and `set_value` are needed. This means if one made a Qt widget and magicgui interface like
  ``` python
  class QMyWidget(QBaseValueWidget): ...

  class MyWidget(ValueWidget):
      def __init__(self, **kwargs):
          kwargs["widget_type"] = QMyWidget
          super().__init__(**kwargs)
  ```
  then it fails with TypeError due to undefined abstract methods.

Please let me know what you think.


tlambert03:
this looks like a great start @hanjinliu, the one thing I'm not thrilled about is the changing of the meaning of `ValueWidget` to `PrimitiveValueWidget`... it's a breaking change to expect everyone to now inherit from `PrimitiveValueWidget`, and as you can see it breaks your own magic-class tests here.  I'd rather use a shared base class and have it not be the case that `isinstance(TupleWidget(), ValueWidget)` than to rename all ValueWidgets

hanjinliu:
That makes sense. Keeping `ValueWidget` in the same state is definitely safer.
The magic-class tests failed by the isinstance check of ValueWidget but I think this should be fixed on my side (this is due to the update in `EmpytWidget`).
The recent breaking change in `griffe` affects the docs test.

tlambert03:
sorry that this has sat so long @hanjinliu.  would you mind resolving conflicts now that TypeMap is in?

tlambert03:
docs are not building correctly anymore.  I think it may require `magicgui.widgets.bases.BaseValueWidget` to be added to the autosummary in docs/api/widgets/bases.md  Please also check whether the mermaid diagram in that document is still correct after this change, new things may need to be added and the connections may need to be changed

hanjinliu:
Thank you for your help! I found that after the implementation of `TypeMap`, functions that refer to the global type map are not very informative (please look at the API reference of `magicgui.type_map.get_widget_class` in the latest docs for instance). It'll be better to have some links pointing to the `TypeMap` method like `(see [TypeMap.get_widget_class](...))`, but we need additional hooks to this because these lines are built by the pre-defined autoref hook.

tlambert03 (approved):
thanks @hanjinliu!

PR #664, opened 2024-08-05
chore: Add missed contents permission to deploy steep of workflow

Czaki:
pointed here https://github.com/napari/napari/pull/7130#issuecomment-2268747965


tlambert03:
Thanks

PR #671, opened 2024-09-11
build: drop python 3.8 add py3.13

tlambert03:


tlambert03:
note that I'm removing one test here:

```python
    assert resolve_single_type(Optional['List["numpy.ndarray"]']) == Optional[List[np.ndarray]]
```

`Optional['List[numpy.ndarray]']` will work, but you can't add an arbitrarily nested number of strings

PR #677, opened 2024-11-07
Implement `TypeMap` to enable module specific type map registration

### LINKED ISSUE #676: `WidgetStore` for module specific type-to-widget mapping
`magicgui` has a global map to determine which widget to use for each type. A problem I'm facing now is that if I updated the type map in my module, it will affect all other modules that depend on `magicgui`. More specifically, I want to use `LineEdit` for all the simple types including `int` and `float` for my purpose but I don't want to change the behaviors of others (for example, if `Annotated[int, {"max": 10}]` is used in another module, it stops working if I change the widget for `int` to LineEdit). I think this problem can be solved by using `type_registered` context manager, but here's much simpler solution and I'd like to discuss whether we should have this feature.

In the `type_map` submodule, we can define a `WidgetStore` class that holds a mapping from type to widget and has all the methods useful for the type mapping tasks:

``` python
class WidgetStore:
    def magicgui(self, ...): ...
    def register_type(self, ...): ...
    def match_type(self, ...): ...
    def get_widget_class(self, ...): ...
    ...
```

This does not affect the current usages because we can just define a default global `WidgetStore` instance.

``` python
_WIDGET_STORE = WidgetStore()

magicgui = _WIDGET_STORE.magicgui
register_type = _WIDGET_STORE.register_type
# ... and so on
```

If one wants to use their own rule, they can just create their own widget store.

``` python
my_store = WidgetStore()

my_store.register_type(int, widget_type=MyIntEdit)

@my_store.magicgui
def func(x: int):
    print(x)
```

If you think it's reasonable, I'm happy to send a PR.

hanjinliu:
Closes #676 

(I said `WidgetStore` in the issue but it turned out that there were more things to store so I simply named it `TypeMap` 🥲 )
All the functions that used the global type map before are now referring to the global instance of `TypeMap`.

Example: 

```python
from magicgui.type_map import TypeMap
from magicgui.widgets import Slider
from magicgui import magicgui

mymap = TypeMap()
mymap.register_type(int, widget_type=Slider)

@magicgui
def func(x: int): ...  # <- x is SpinBox, as usual

@mymap.magicgui
def func(x: int): ...  # <- x is Slider

```

Some functions, such as `magicgui` and `create_widget`, no longer need to be in the files they used to be, but they are still in the original files in case some users use private import such as `from magicgui.type_map._magicgui import magicgui`. Probably we need some refactoring in the future because there are some dirty import statements inside functions to avoid circular import.


hanjinliu:
It seems `pint` has some problem with dataclass (see [here](https://github.com/hgrecco/pint/issues/1969)).

hanjinliu:
Ah, that's why all the tests including python<3.13 are failing.

tlambert03:
flexparser is fixed ... could you merge conflicts now that #671  is merged?

hanjinliu:
Is the failed test a random connection error? Everything is working as expected :)

tlambert03 (approved):
this is excellent work @hanjinliu!  thank you so much.

PR #678, opened 2024-11-22
fix: fix safe_issubclass for NewType

Czaki:
Closes https://github.com/napari/napari/issues/7129

Today I learn that such code:

```python 

new_int = NewType("new_int", int)

issubclass(new_int, new_int)
```

Raises exception instead of returning True. 

So this PR is a followup to #665  to fully resolve the upstream napari bug.


tests/test_util.py
```diff
@@ -80,3 +80,12 @@ def test_subclass_future(self):
...
         assert not safe_issubclass(Future[list[int]], Future[list[str]])
+
+    def test_subclass_new_type(self):
+        new_int = typing.NewType("new_int", int)
+
+        assert safe_issubclass(new_int, new_int)
+        assert safe_issubclass(list[new_int], list[new_int])
+        assert safe_issubclass(list[new_int], list[new_int])
+        assert safe_issubclass(list[new_int], typing.Sequence[new_int])
+        assert safe_issubclass(list[new_int], list[new_int])
```
jni, line 91: Why is this assert repeated three times @Czaki? And should one also test typing.List?

Czaki: ruff....

Czaki:
@tlambert03 The #671 changed type annotation used in `tests/test_util.py` making many test repetitive. 
Should we restore these changes or simplify tests?

tests/test_util.py
jni: 🤣 that is amazing

src/magicgui/_util.py
```diff
@@ -187,6 +187,8 @@ def _safe_isinstance_tuple(obj: object, superclass: object) -> bool:
 
 def safe_issubclass(obj: object, superclass: object) -> bool:
     """Safely check if obj is a subclass of superclass."""
+    if obj == superclass:
```
tlambert03, line 190: when comparing types, shouldn't we use `is` ?

tlambert03:
> Should we restore these changes or simplify tests?

ah, i think these should probably be restored.  since `typing.List` is still not an alias for `list` (in fact they don't even equal each other), it probably doesn't hurt to just keep checking that the typing objects work explicitly

src/magicgui/_util.py
Czaki:
```suggestion
    if obj is superclass:
```
lets check

tlambert03:
Thanks!

tlambert03:
Do you need this released asap?  Can it wait a week or so to give time for testing the current pre-release?  Or would you prefer a cherry pick back-release

Czaki:
This is a question to @psobolewskiPhD

psobolewskiPhD:
I mean it's been around and it only affects things that didn't update to drop older python, so I don't think it's critical?
I think being able to update the image.sc thread that it's been fixed and say look for a new magicgui release in a week or 2 should be fine?

tlambert03:
Sounds good thanks

PR #679, opened 2024-11-23
chore: restore types in tests/test_util.py

Czaki:
https://github.com/pyapp-kit/magicgui/pull/678#issuecomment-2494967068

As `typing.List` is not same as `list` we need to revert this change introduced in #671


tlambert03:
Thanks!

PR #688, opened 2025-02-02
fix: fix name collisions between dataclass fields and guiclass container attributes

### LINKED ISSUE #684: `guiclass` and the `name` reserved keyword
**Describe the bug**
If your data model has a property called `name` the example provided on the website crashes

**To Reproduce**
```python
from magicgui.experimental import guiclass, button

@guiclass
class MyDataclass:
    a: int = 0
    name: str = 'hello'
    c: bool = True

    @button
    def compute(self):
        print(self.a, self.b, self.c)

obj = MyDataclass(a=10, name='foo')
obj.gui.show()
```
**Expected behavior**
The example should run if the property is called `name`

**Screenshots**

 - OS: Win
 - backend:  QT
 - magicgui version 0.10.0

tlambert03:
fixes #684

When a container has a widget whose `name` attribute matches an attribute on the `Widget` class itself, then the overloaded `__getattr__` method becomes a problem.  (reminder: that getattr is what makes it possible to retrieve a magicgui function parameter (for example) as an attribute on the magicgui container widget).

Here I add a `get_widget(self, name:str)` method on Container that is *guaranteed* to either return a widget instance or None, avoiding conflicts with other attributes no the Widget instance itself


tlambert03:
@hanjinliu, could you have a look at the magic-class error and see what i've done here that is selectively breaking magic-class?  thanks!

hanjinliu:
Hi @tlambert03 , I found that the older version and this version returns inconsistent result when a Container widget has multiple widgets with the same name.

```python
from magicgui import __version__
from magicgui.widgets import Container, LineEdit
c = Container()
c.append(LineEdit(name="a", value="1"))
c.append(LineEdit(name="a", value="2"))
c["a"].value
```

This code returns `1` in the older version, but returns `2` for this PR. The `magicclass` error occurred because there are widgets of different types.
In the order version, `__getattr__` (and `__getitem__`) searches for the widget whose name matches the input `str` by for-loop from the beginning of `_list`. Maybe something changed in this PR?

tlambert03:
ok, that makes sense... it's caused by this change:

https://github.com/tlambert03/magicgui/blob/8deddf6c7e93cb160ca5dbb06ded2e666a2160c5/src/magicgui/widgets/bases/_named_list.py#L26-L36

I'd say that the implicit expectation of getting the first matching widget in the case of a name collision rather than the last matching widget was probably always more bug-like and not any sort of formal protocol here, but if that's all it takes not to break magic-class I can go back to that behavior.  Perhaps after this PR, you can also have a look at places where you have behavior (either explicit or implicit) that might change depending on whether someone has added two widgets with the same `name` attribute.  I don't think either of us should depend on name lookup in that case... 

In the case of a dataclass (which precipitated this PR), i think it *is* safe to assume that there will never be two widgets with the same name, but that is probably a unique case

hanjinliu:
Yes, I definitely have to avoid adding widget of the same name on my side.
I think that adding widgets of the same name is always dangerous and it is a good idea to encourage people to make widgets from dataclass, TypedDict, pydantic BaseModel etc. to prevent collision errors at the language level.

PR #700, opened 2025-06-03
build: use uv for project management

tlambert03:
out with the old, in with the new.  ❤️ uv


pyproject.toml
```diff
@@ -31,117 +33,102 @@ classifiers = [
...
-]
-
 
 # https://docs.astral.sh/ruff
 [tool.ruff]
 line-length = 88
 target-version = "py39"
 src = ["src", "tests"]
+fix = true
+unsafe-fixes = true
```
Czaki, line 131: Brave. may be confusing for contributors, as may break the code.

uv.lock
Czaki, line 1: As you add uv.lock to the repository, what is your plan to refresh it? 
As I remeber, the uv run will be use versions from uv.lock file.

tlambert03: can omit.  will do

PR #669, opened 2024-09-06
Nested pydantic model containers

ndxmrb:
Hi @tlambert03,
when using nested pydantic models, I found [this post](https://forum.image.sc/t/building-a-napari-widget-from-a-pydantic-model/90257/11) on how to add a simple way of nesting widgets/containers. When trying to use the values, I needed to modify `asdict()` as well. Although there might be a reason you didn't just implement it already when answering the post (probably due to #317), I thought I'd make a pull request just in case.

Notably, some things are missing:
* Proper test e.g. for getting nested values from `ValueWidget`s and `ContainerWidget`s - as I didn't really know where to start
* QGroupBox option as mentioned in #317 

Also, I tried my best with the types, but I probably don't oversee the grand scheme of things... 😅 

I hope it's helpful, but I'm aware it might not be, so feel free to just close 😄 

Thanks for having a look!

Issue #317, opened 2021-11-13
Improve nesting of containers with labels (group box)


aeisenbarth:
Sometimes you want to structure user interfaces by grouping widgets. Currently when adding two containers A and B to a Napari dock (a parent container), the widgets in the nested containers are displayed in a continuous flow and the grouping is not recognizable by users.

A possibility is to set a label for A and B and enable label display on the parent container. However, as with normal widgets, Container labels are displayed on the left, squeezing the container content to the right. In this example, Segmentation is a container with vertical layout for which I would like to have the container's label displayed above its content and the container separated by a border from other widgets. Shape is also a container (horizontal layout), as an example that the current behavior can also be desirable to keep.

<img src="https://user-images.githubusercontent.com/54448967/141651065-dc1c72a6-2c69-44b1-b8ec-d27f0ac97707.png" width=50% height=50%>

It would be nice to have something like [QGroupBox](https://doc.qt.io/qt-5/qgroupbox.html) which could either be its own magicgui widget or an option for Container.

michaels10:
Came up for us too. Any news on fixing this?

ndxmrb:
Hi,
I'd love to have this feature as well. 

In case anyone using pydantic models is wondering how to get what is shown above in the first place, please refer to [this post](https://forum.image.sc/t/building-a-napari-widget-from-a-pydantic-model/90257/11) and/or #669.

PR #704, opened 2025-06-19
Nested containers

ndxmrb:
Hi @tlambert03,
restarting my work from #669 after the `ValuedContainer` changes.

Still, some things are missing:
* Proper test e.g. for getting nested values from `ValueWidget`s and `ContainerWidget`s. Does this belong to `test_pydantic()`in `test_ui_field.py`?
* QGroupBox option as mentioned in #317 

Tried my best with the types again, hope it's okay that way.

Thanks for having a look!


tlambert03 (commented):
hey @ndxmrb 

so sorry for the long delay.  couple concerns about typing and error catching.  we don't need to belabor them too much, if you can fix great, otherwise let me know.

ndxmrb (commented):
Happy new year, @tlambert03!
although I forgot, I haven't lost interest in any of this. Creating the concrete class is indeed the missing piece. I remember I had hoped `ValuedContainerWidget` would be that already. I guess I bit off more than I could chew. Thanks for picking it up.

PR #724, opened 2026-04-10
fix: use default font from application instead of system default

giocard:
This is a change proposed to solve the warning issue experienced on Windows with napari. See https://github.com/napari/napari/issues/8877 for a complete description of it.


tlambert03 (approved):
great, thanks!

PR #727, opened 2026-05-11
chore: Use `dependency-group` for napari in ci test

TimMonko:
napari/napari#8704

Uses a matrix value or blank for either `dependency-extras` or `dependency-group` in the test_dependents ci.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -141,11 +141,12 @@ jobs:
     secrets: inherit
 
   test-dependents:
-    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/test_and_deploy.yml
```diff
@@ -141,25 +141,30 @@ jobs:
...
-      python-version: "3.10"
+      python-version: ${{ matrix.python-version || '3.10' }}
       post-install-cmd: "python -m pip install pytest-pretty lxml_html_clean" # just for napari
       pytest-args: ${{ matrix.pytest-args }}
     strategy:
       fail-fast: false
       matrix:
         include:
           - package: napari/napari
+            python-version: "3.13"
```
tlambert03, line 159: unless we want to actually tests multiple versions, go ahead and just change the python-version above directly to '3.13'.  it's a once-per-3-year change

PR #736, opened 2026-07-24
fix: Table.data support for negative indices

### LINKED ISSUE #735: Table data get item doesn't work with negative indices
**Describe the bug**
Getting or setting table data with slice notation, e.g. `table.data[0, 0]`, doesn't work with negative indices.

I could probably do a PR, but I am not sure if the fix should be implemented at the `widgets.Table` level, or lower.

**To Reproduce**

```python
from magicgui import widgets

tbl = widgets.Table([[1, 2, 3], [4, 5, 6]])

print(f"Value at [0, 2]: {tbl.data[0, 2]}")    # Value at [0, 2]: 3
print(f"Value at [0, -1]: {tbl.data[0, -1]}")  # Value at [0, -1]: None
```

**Expected behavior**

I would expect `tbl.data[0, -1]` to be function like all other negative indices in Python.

At the very least, if they don't work, supplying a negative index should raise an `IndexError` (although I think most people would find this surprising)

**Screenshots**
N/A

 - OS: Windows 11
 - backend: PyQt6 v6.9.1
 - magicgui version: 0.10.3.dev3+g2d985e7a4

Antyos:
Qt native Table `_mgui_get_cell()` and `_mgui_set_cell()` now support negative indices and will raise an `IndexError` if invalid indices are specified.

Previously, getting or setting to an invalid index would do nothing. Raising an `IndexError` is probably more useful, but may break some existing code.

Fixes #735.

### DIFF
--- a/src/magicgui/backends/_qtpy/widgets.py
+++ b/src/magicgui/backends/_qtpy/widgets.py
@@ -1520,13 +1520,20 @@ class Table(QBaseWidget, protocols.TableWidgetProtocol):
             normalized_row = row + num_rows
         else:
             normalized_row = row
-        if col < 0: 
+        if col < 0:
             normalized_col = col + num_cols
         else:
             normalized_col = col
-        # Qt will not raise an error for attempting to index into an invalid column, but normal Pyhton does.
-        if normalized_row >= num_rows or normalized_col >= num_cols:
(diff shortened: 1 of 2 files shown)


tlambert03:
thanks @Antyos!  test fails seem unrelated, I'll clean them up in a followup

PR #737, opened 2026-07-26
fix: make MagicFactory work as a method descriptor

### LINKED ISSUE #733: `@magic_factory` doesn't work as a method descriptor properly
**Describe the bug**
Using `@magic_factory` on a class method will not properly pass `self` to its generated function guis, likely because it is using `functools.partial` instead of `functools.partialmethod` internally.

**To Reproduce**

```python
from magicgui import magic_factory


class C:
    def __init__(self, x: int):
        self.x = x
        self.func = self.func_fact()

    @magic_factory()
    def func_fact(self):
        print(self.x)

    # In in ideal world, this would also work:
    # func = func_fact()


c = C(x=5)

c.func()
# c.func_fact()()  # This also fails

```

<details><summary>Traceback</summary>
<p>

```python
c:\Users\andrew\Code\project\code\sandbox\magic_factory_method.py:7: FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior
  self.func = self.func_fact()
  File "c:\Users\andrew\Code\project\code\sandbox\magic_factory_method.py", line 16, in <module>
    c.func()
    ~~~~~~^^
  File "C:\Users\andrew\Code\project\code\.venv\Lib\site-packages\magicgui\widgets\_function_gui.py", line 338, in __call__
    raise TypeError(msg) from None
TypeError: missing a required argument: 'self' in call to 'func_fact(self)'.
To avoid this error, you can bind a value or callback to the parameter:

    func_fact.self.bind(value)

Or use the 'bind' option in the magicgui decorator:

    @magicgui(self={'bind': value})
    def func_fact(self): ...
```

</p>
</details>

**Expected behavior**
I would expect `@magic_factory` to work on a class method.

 - OS: Windows, Python 3.13.9
 - backend: PyQt6==6.9.1
 - magicgui version: from main branch

Sanjays2402:
Closes #733

`functools.partial` had no `__get__` before Python 3.13, so a `@magic_factory` used on a method never received the instance; from 3.13 on `partial.__get__` binds the instance as the first *positional* argument, which `MagicFactory.__call__` rejects. Either way calling the resulting widget raised `TypeError: missing a required argument: 'self'`.

`MagicFactory` now implements `__get__` itself and returns a factory that binds its first parameter to the instance, mirroring `FunctionGui.__get__`. New `tests/test_factory.py::test_magic_factory_as_method` fails without the fix and passes with it.


tlambert03:
thanks for the fix!  Sorry for the dealy

PR #738, opened 2026-07-26
fix: bind magic factories as method descriptors

### LINKED ISSUE #733: `@magic_factory` doesn't work as a method descriptor properly

JosephDemarest:
## Summary
- bind the decorated function when a `MagicFactory` is accessed through an instance
- preserve class-level access, factory options, `widget_init`, and the originating type map
- cover independent instance binding and ensure a temporary bound factory does not retain its instance

Fixes #733

## Tests
- Python 3.12.9: `python -m pytest -q` (`330 passed, 51 skipped, 13 xfailed`)
- Python 3.14.6: `python -m pytest tests/test_factory.py -q` (`12 passed`)
- `pre-commit run --all-files`
- `mypy src/magicgui/type_map/_magicgui.py`

### DIFF
--- a/src/magicgui/type_map/_magicgui.py
+++ b/src/magicgui/type_map/_magicgui.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import inspect
 from functools import partial
-from typing import Any, Callable, Generic, TypeVar
+from typing import Any, Callable, Generic, TypeVar, cast
 
 from magicgui.type_map._type_map import TypeMap
 from magicgui.widgets import FunctionGui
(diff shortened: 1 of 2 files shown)


tlambert03:
Thank you for the PR @JosephDemarest, this was preceded by another PR that fixes the same issue #737 ... which was just merged.  Thanks again

PR #741, opened 2026-08-24
feat: more complete main window implementation

### LINKED ISSUE #601: MainWindow for ipywidgets
the Qt backend has a MainWindow but the ipywidgets backend doesn't.  As @larsoner has pointed out, the closest ipywidgets analog is probably [`AppLayout`](https://ipywidgets.readthedocs.io/en/7.x/examples/Layout%20Templates.html#AppLayout)

In Qt, a [main window](https://doc.qt.io/qt-6/qmainwindow.html#qt-main-window-framework) is characterized by the ability to add a menu bar (top), status bar (bottom), dock widgets, and toolbars:

mainwindowlayout

In ipywidgets, it's more a layout issue, with a header, a footer, two sidebars and a central pane:

Screen Shot 2023-10-11 at 5 24 57 PM

Currently, our MainWindow protocol is a minimal subclass of `Container` that can add menus:

```python
class MainWindowProtocol(ContainerProtocol, Protocol):
    def _mgui_create_menu_item(
        self,
        menu_name: str,
        action_name: str,
        callback: Callable | None = None,
        shortcut: str | None = None,
    ) -> None:
```

but will likely gain `_mgui_add_toolbar` after #597...

This raises the question of how represent these somewhat different concepts of "main window" in our API.  If we use `AppLayout` behind the scenes, how do we assign stuff to left/middle/right?  Or do we just ignore the left and right (and allow people to manually edit using `widget.native` if they choose.  

fwiw, it looks like mne-python [doesn't use AppLayout](https://github.com/mne-tools/mne-python/blob/main/mne/viz/backends/_notebook.py) directly anywhere?

@larsoner, any insights/opinions from your mne applications?

larsoner:
Closes #601
Closes https://github.com/pyapp-kit/magicgui/pull/604

So... the "next week" I promised in 2023 has finally arrived 😅 This PR has the 43 commits from that PR, plus what Claude Fable found while reviewing (below is its summary). I reviewed the commits and they make sense to me. Notably I removed the backend autodetection since we should probably scope / implement that separately. (Automatic IPython importing can [be problematic](https://github.com/mne-tools/mne-python/blob/acb46e2bca5b9887705d53336f11d594e8e9b63d/doc/sphinxext/mne_doc_utils.py#L180-L191), and maybe it should be opt-in, etc.)

Before the Fable fixes:

mw_shot

After:

mw_shot4

Qt (I'm on macOS so the menu isn't captured in-window):

pr604_qt

---

- Merged with current `main` (~3 yrs of drift)
- Made the ipynb `MainWindow` work as a real `Container`: children now flow into the grid's central cell (previously `main.append(...)` raised `KeyError`).
- Implemented ipynb `MenuBar`/`Menu` as a row of Dropdowns per the #601 discussion (first entry is the title placeholder; selecting an entry fires its callback and resets); nested submenus raise `NotImplementedError` for now.
- Fixed grid placement: each bar/dock box gets its own `Layout` instance (they shared   one, so `GridspecLayout.__setitem__` stacked all eight into a single cell), and the   grid templates are set via the attributes `_update_layout()` re-applies, so empty  areas collapse and the central widget gets `1fr`.
- ipynb `StatusBar`: added widgets are tracked separately from the message label and spacer, negative insert positions append (matching Qt), and message timeouts use `get_running_loop()` instead of the deprecated `get_event_loop()`.
- `set_menu_bar(None)` / `set_status_bar(None)` no longer crash in the ipynb backend.
- Drive-by fix for a pre-existing bug on `main`: Qt `Container`/`Dialog` reported `layout == "vertical"` regardless of orientation (`isinstance(self, QHBoxLayout)` can never be true for a backend widget).
- Exported `Menu`/`MenuBar` from `magicgui.widgets`, and fixed a `raise self._ipywidget` → `return` typo in `_mgui_get_native_layout`.
- Added backend-parametrized tests for the dock/toolbar/status-bar/menu APIs (should address the codecov failure).
- Moved the root `example.py` into `docs/examples/applications/main_window.py`.
- Dropped the `_choose_backend()` jupyter auto-detection for now — it's separable and changes default-backend behavior for all users, so it can be its own PR.


larsoner:
Okay merged `main` in to resolve conflicts and pushed one last test fix (locally things were failing... had to do with not completely un-setting states

tlambert03:
Thanks a lot for coming back to this @larsoner, I'm more or less happy to merge this.  Strictly new feature, shouldn't affect any existing behavior (unless someone is creating a concrete `MainWindowProtocol` outside of magicgui, which I don't expect).

local claude review did surface these two small bugs... do you agree that they should be fixed?  Or is it wrong?

- **Qt `MenuBar._mgui_clear` is an empty stub** (docstring only, no body) — src/magicgui/backends/_qtpy/widgets.py. So `main.menu_bar.clear()` silently does nothing on Qt while it works on ipynb.  One-line fix : `self._qwidget.clear()`.
- **`_SupportsMenus._menus` is never cleared**, so `menu_bar["File"]` still returns the stale menu after `clear()`.

larsoner:
Claude was right! Fixes pushed

tlambert03:
thanks!

PR #743, opened 2026-08-24
test: deterministically tear down example widgets

larsoner:
CIs seem to be failing in all PRs. Based on (painful) experience with PyVistaQt, mne-qt-browser, etc. a reasonable fix is to make sure windows are closed and torn down at the end of every test (rather than "whenever gc.collect() decides to run"). Otherwise, a still-visible widget whose C++ side is deleted mid-event-loop can receive a paintEvent during deletion, which leads to a segfault.

This should hopefully fix CIs. Changes drafted by Claude Fable 5 but reviewed / iterated / understood be me.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,15 @@ build-backend = "hatchling.build"
 [tool.hatch.version]
 source = "vcs"
 
+# hatchling 1.32 emits core metadata 2.5 by default, which twine (as run by
+# build-and-inspect-python-package) rejects.  TODO: remove once twine there
+# accepts metadata 2.5.
+[tool.hatch.build.targets.sdist]
+core-metadata-version = "2.4"
+
+[tool.hatch.build.targets.wheel]
+core-metadata-version = "2.4"
+
 # https://peps.python.org/pep-0621/
 [project]
 name = "magicgui"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,11 @@ third-party-support = [
     "toolz>=1.0.0",
 ]
 test-min = ["pytest>=8.4.0", "pytest-cov >=6.1", "pytest-mypy-plugins>=3.1"]
-test-qt = [{ include-group = "test-min" }, "pytest-qt ==4.4.0"]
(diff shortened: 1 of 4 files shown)


larsoner:
... let me know if you want me to add a "open an issue when the scheduled one fails, if one doesn't already exist" here or in another PR. I find them useful in other repos, otherwise I miss scheduled run failures (but not sure if it would help here, depends on maintainer availability etc.)

tlambert03:
Rebased this onto current `main` and dropped the napari commit (23cc0616), which #727 superseded this morning — that job now runs on 3.13 with `package-group: testing` and is passing. That was also the only merge conflict, so the branch is clean now.

Confirmed your pytest-qt diagnosis independently before touching anything: the failure is `_AbstractSignalBlocker._cleanup` calling `_silent_disconnect(self._timer.timeout, ...)`, which only catches `(TypeError, RuntimeError)`. PySide6 6.11.2 raises `SystemError` there instead, so it escapes inside a psygnal callback → `EmitLoopError` → `waitSignal` never returns → `TimeoutError`. pytest-qt 4.5.0 removed that disconnect from the cleanup path entirely, so the bug can't trigger.

Verified locally with `pyside6==6.11.2` (what CI resolves; our lock had been pinning 6.11.0, which is why this reproduced only in CI): 423 passed. PyQt5/PyQt6 also green, and the `conflicts` split resolves correctly — `pyside2` → `pytest-qt==4.4.0`, everything else → `4.5.0`.

I left the hatchling core-metadata commit in. Worth noting it hasn't actually been exercised yet: `build-and-inspect-package` has `needs: test`, so it's been skipped this whole time. We'll find out on this run.

Thanks for chasing this one down.

larsoner:
You still have 3.9 tests... want to bump to 3.10 (or if you're bold, 3.11)? Happy to do this in another maintainability PR

tlambert03:
i can bump min python elsewhere.  thanks.

and as for the "open an issue when the scheduled one fails", I too use/like that elsewhere, but i fear I'm the only one listening here at the moment, and I'm feeling pretty buried recently (and likely wouldn't take action on it until i had to anyway)

PR #745, opened 2026-08-26
build!: drop python 3.9 and 3.10, drop pyside2, require >=3.11

tlambert03:
Drops Python 3.9/3.10 and moves the floor to `>=3.11`, plus the ruff updates that fall out of it.

## ⚠️ This also drops PySide2

Not a choice so much as a consequence: PySide2 `5.15.2.1` declares `requires_python: ... <3.11` and ships no wheel past `cp310`. It cannot be installed on the new floor, so its extra, dependency-group and CI jobs are all dead weight. Removing it also lets us delete the `test-qt-legacy` group and the `[tool.uv] conflicts` block from #743, which existed *only* to hold PySide2 on pytest-qt 4.4.

## Changes

**Version floor**
- `requires-python = ">=3.11"`, 3.9/3.10 classifiers removed
- CI: base matrix `3.10`→`3.11`, min-deps `3.9`→`3.11`, pyqt5 job `3.9`→`3.11`, `test-dependents` default `3.10`→`3.11` (it would otherwise try to install magicgui on 3.10)
- `SLOTS` is unconditional now; the `<3.11` `weakref_slot` skip is dead code
- PyQt5 is untouched — it supports 3.11+, only its job's Python moved

**Ruff**
- `target-version` `py39`→`py311`
- Dev toolchain re-pinned to what pre-commit already runs. These had drifted badly: `ruff>=0.8.3` resolved to **0.15.12** locally while pre-commit pinned **0.16.4** — different enough that `RUF036` is preview in one and stable in the other, so `uv run ruff` and CI disagreed about the codebase.
- 91 autofixes: `Callable` → `collections.abc`, `typing_extensions` → `typing` where 3.11 suffices, `zip(..., strict=False)` (explicit, behavior-preserving), one dead `type: ignore`

## Two rules deliberately disabled

**`UP007`/`UP045`** (`Optional[X]` → `X | None`) are in `ignore`, because for this codebase they are not a cosmetic change:

```python
get_origin(Union[int, None]) is Union   # True
get_origin(int | None) is Union         # False on 3.11-3.13, True on 3.14
```

magicgui compares `get_origin(...) is Union` in four places, so PEP 604 unions take a different path. Letting ruff apply these broke 7 tests in `test_ui_field.py` — it rewrote `Optional[int]` to `int | None` and `UiField` then extracted a different annotation. Tests also deliberately exercise both spellings, so blanket-rewriting them would quietly cost coverage.

**`TC003`** — `collections.abc` is added to `flake8-type-checking.exempt-modules`, alongside the `typing`/`typing_extensions` that ruff exempts by default. These 16 findings only appeared because `Callable` moved off `typing`; magicgui resolves annotations at runtime, so those names must stay importable at runtime rather than move into `TYPE_CHECKING`.

## Known pre-existing bug (not fixed here)

The above uncovered a real bug on `main`, unrelated to this PR:

```python
@magicgui
def f(x: Optional[int] = None): ...   # .annotation -> int          ✅

@magicgui
def f(x: int | None = None): ...      # .annotation -> int | None   ❌
```

PEP 604 unions aren't unwrapped by the four `get_origin(...) is Union` checks on Python 3.11–3.13 (masked on 3.14, where `types.UnionType` became `typing.Union`). Worth its own PR; happy to write it.

## Testing

Full suite green on **3.11 / 3.13 / 3.14**, across **PyQt6, PyQt5, PySide6**. `prek run -a` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.11", "3.13"]
         os: [ubuntu-latest, macos-latest, windows-latest]
         add-group: [pyqt6, pyside6]
         include:
@@ -31,18 +31,12 @@ jobs:
             python-version: "3.11"
             resolution: "lowest-direct"
             add-group: pyside6
-          - python-version: "3.9"
+          - python-version: "3.11"
             os: ubuntu-latest
             add-group: pyqt5
-          - python-version: "3.9"
-            os: ubuntu-latest
-            add-group: pyside2
           - python-version: "3.12"
             os: windows-latest
             add-group: pyqt5
-          - python-version: "3.10"
-            os: windows-latest
-            add-group: pyside2
           - python-version: "3.12"
             os: ubuntu-latest
             add-group: pyqt6
@@ -109,7 +103,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.13"]
+        python-version: ["3.11", "3.13"]
         os: [macos-latest]
 
     steps:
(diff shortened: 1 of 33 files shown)


larsoner (approved):
I dropped PySide2 in PyVistaQt as well, no regrets! 😄

